PR #12, opened 2019-05-13
Feature/hw 52030 user repository

fmattos-hw:


fmattos-hw:
This PR contains the change to consider the user profile type when creating a transfer method

PR #19, opened 2019-05-15
HW-52170 key and field gql integration

peter-joseph:


ui/src/main/java/com/hyperwallet/android/ui/transfermethod/SelectTransferMethodPresenter.java
```diff
@@ -15,23 +15,24 @@
...
  * IN THE SOFTWARE.
  */
-
 package com.hyperwallet.android.ui.transfermethod;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 import com.hyperwallet.android.model.HyperwalletErrors;
 import com.hyperwallet.android.model.HyperwalletUser;
```
fmattos-hw: This class has a few inconsistencies with regards to variable names. In some methods, we have String country and others String countryCode. Could you please make the variable names consistent in the whole class. Same applies to other variables

ui/src/main/java/com/hyperwallet/android/ui/transfermethod/SelectTransferMethodPresenter.java
```diff
@@ -110,24 +109,24 @@ public void loadCurrency(final boolean forceUpdate, @NonNull final String countr
...
-                                    @Nullable final HyperwalletTransferMethodConfigurationKeyResult result) {
+                            public void onKeysLoaded(@Nullable final HyperwalletTransferMethodConfigurationKey key) {
                                 if (!mView.isActive()) {
                                     return;
                                 }
-                                List<String> transferMethodCurrencies = result.getCurrencies(countryCode);
-                                List<String> transferMethodTypes = result.getTransferMethods(
-                                        countryCode, transferMethodCurrencies.get(0), user.getProfileType());
+                                List<com.hyperwallet.android.model.meta.keyed.Currency> transferMethodCurrencies =
+                                        new ArrayList<>(key.getCurrencies(countryCode));
```
fmattos-hw: is it possible `key.getCurrencies(countryCode)` returning null? If so we could have a npe here

PR #23, opened 2019-05-18
HW-51714 - Updating gql responses

skoong:


ui/src/androidTest/java/com/hyperwallet/android/transfermethod/ui/PayPalTest.java
```diff
@@ -126,10 +126,10 @@ public void testAddTransferMethod_displaysFeeElementsOnTmcResponse() {
 
         onView(withId(R.id.add_transfer_method_fee_label)).check(
                 matches(withText(R.string.add_transfer_method_fee_label)));
-        onView(withId(R.id.add_transfer_method_processing_label)).check(
-                matches(withText(R.string.add_transfer_method_processing_time_label)));
+//        onView(withId(R.id.add_transfer_method_processing_label)).check(
```
fmattos-hw: It would be a good idea adding a "todo" here as it's done in BankAccountTest.java. Also in BankCardTest.java (same test case, testAddTransferMethod_displaysFeeElementsOnTmcResponse) I don't see any changes. Is that test going to fail too, correct? If not I think we missed something there

skoong: Thanks, updated.  If a view is has visibility GONE, the value can still be checked and the test will not fail.  Added an additional matcher for visibility.

PR #25, opened 2019-05-22
HW-52174. Added datePicker for DateWidget

azakrevska-epam:
Created new branch and PR due to problems with rebase and merged parent branches


ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateUtil.java
```diff
@@ -0,0 +1,134 @@
...
+    @NonNull
+    String convertDateFromServerToWidgetFormat(@Nullable final String serverDate) {
+        if (isServerDateNotValid(serverDate)) {
+            return "";
+        }
+        try {
+            final Calendar calendar = Calendar.getInstance();
+            calendar.setTime(mServerDateFormat.parse(serverDate));
+            return mWidgetDateFormat.format(calendar.getTime());
+        } catch (NumberFormatException | ParseException | IndexOutOfBoundsException e) {
```
peter-joseph: if we are really not concern about the exception why don't we throw it back? or why not do the following
```
    String convertDateFromServerToWidgetFormat(@Nullable final String serverDate) {
        String converted = "";
        if (isServerDateValid(serverDate)) {
            try {
                final Calendar calendar = Calendar.getInstance();
                calendar.setTime(mServerDateFormat.parse(serverDate));
                converted = mWidgetDateFormat.format(calendar.getTime());
            } catch (Exception e) { // exception will not be processed and defaults this process to empty string}
        }
        return converted;
    }

File: ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateUtil.java [hunk 2]
```diff
@@ -0,0 +1,134 @@
...
+    Calendar convertDateFromServerFormatToCalendar(@Nullable final String serverDate) {
+        final Calendar calendar = Calendar.getInstance();
+        if (isServerDateNotValid(serverDate)) {
+            return calendar;
+        }
+
+        try {
+            calendar.setTime(mServerDateFormat.parse(serverDate));
+            return calendar;
+        } catch (ParseException e) {
```
peter-joseph: if we don't care about the exception why not throw it back to the caller maybe it will make more sense to process that there? if not why not clean the code like so
```
    Calendar convertDateFromServerFormatToCalendar(@Nullable final String serverDate) {
        final Calendar calendar = Calendar.getInstance();
        if (isServerDateValid(serverDate)) {
            try {
                calendar.setTime(mServerDateFormat.parse(serverDate));
            } catch (ParseException e) { // will not process exception and defaults calendar to its own current instances' time and date }
        }
        return calendar;
    }

File: ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateUtil.java [hunk 3]
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    private boolean isServerDateNotValid(@Nullable final String serverDate) {
+        Date date = null;
+        if (!TextUtils.isEmpty(serverDate)) {
+            SimpleDateFormat formatter = new SimpleDateFormat(SERVER_DATE_PATTERN, Locale.getDefault());
+            formatter.setLenient(Boolean.FALSE); // make it strict
+            try {
+                date = formatter.parse(serverDate);
+            } catch (ParseException e) {
```
peter-joseph: Exception processing same here

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateUtil.java
```diff
@@ -0,0 +1,134 @@
...
+
+        try {
+            calendar.setTime(mServerDateFormat.parse(serverDate));
+            return calendar;
+        } catch (ParseException e) {
+            return calendar;
+        }
+    }
+
+    private boolean isServerDateNotValid(@Nullable final String serverDate) {
```
peter-joseph: can we change this to more positive check so that it will be easier to read? `isServerDateValid` and change return to `return date != null` ?

ui/src/main/java/com/hyperwallet/android/ui/transfermethod/AddTransferMethodFragment.java
```diff
@@ -580,4 +584,30 @@ private boolean performValidation(boolean bypassFocusCheck) {
     public boolean isActive() {
         return isAdded();
     }
+
+    @Override
+    public void openWidgetDateDialog(final int year, final int month, final int dayOfMonth,
+            @NonNull final DateChangedListener dateChangedListener) {
+        DatePickerDialog datePickerDialog = new DatePickerDialog(requireContext(),
```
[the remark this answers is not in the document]
azakrevska-epam: @peter-joseph Thank you, added

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateWidget.java
```diff
@@ -51,52 +53,41 @@ public View getView(@NonNull final ViewGroup viewGroup) {
...
 
             // input control
             mTextInputLayout = new TextInputLayout(new ContextThemeWrapper(viewGroup.getContext(),
                     mField.isEditable() ? R.style.Widget_Hyperwallet_TextInputLayout
                             : R.style.Widget_Hyperwallet_TextInputLayout_Disabled));
-            final EditText editText = new EditText(
+
+            mEditText = new EditText(
                     new ContextThemeWrapper(viewGroup.getContext(), R.style.Widget_Hyperwallet_TextInputEditText));
+            if (!TextUtils.isEmpty(mField.getValue())) {
```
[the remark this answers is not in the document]
azakrevska-epam: @peter-joseph Thank you. You are right. Fixed.

ui/src/main/java/com/hyperwallet/android/ui/transfermethod/AddTransferMethodFragment.java
```diff
@@ -580,4 +584,30 @@ private boolean performValidation(boolean bypassFocusCheck) {
...
+                new DatePickerDialog.OnDateSetListener() {
```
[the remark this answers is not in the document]
azakrevska-epam: @peter-joseph Thank you. You are right. Fixed.

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateUtil.java
azakrevska-epam: Done

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateUtil.java
azakrevska-epam: @peter-joseph I added custom Exception to deal with incorrect result in the widget code

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateUtil.java
azakrevska-epam: Added DateParseException

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateUtil.java
azakrevska-epam: Created DateParseException

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateDialogFragment.java
```diff
@@ -0,0 +1,118 @@
...
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.fragment.app.DialogFragment;
+import androidx.fragment.app.FragmentManager;
+
+import com.hyperwallet.android.hyperwallet_ui.R;
+
+import java.util.Calendar;
+
+public class DateDialogFragment extends DialogFragment {
```
peter-joseph: Please move this class to package `com.hyperwallet.android.ui.view`

ui/src/main/java/com/hyperwallet/android/ui/transfermethod/AddTransferMethodActivity.java
```diff
@@ -190,4 +193,29 @@ private AddTransferMethodFragment getAddTransferMethodFragment() {
         }
         return fragment;
     }
+
+    @Override
+    public void showDateDialog(@Nullable final String date, @NonNull final String fieldName) {
```
peter-joseph: Please put this method inside `AddTransferMethodFragment`

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateDialogFragment.java
```diff
@@ -0,0 +1,118 @@
...
+                getString(R.string.cancel_button_label),
+                new DialogInterface.OnClickListener() {
+                    @Override
+                    public void onClick(DialogInterface dialog, int which) {
+                        if (which == DialogInterface.BUTTON_NEGATIVE) {
+                            mOnSelectedDateCallback.setSelectedDateField(fieldName, null);
+                        }
+                    }
+                });
+
```
peter-joseph: The dialog will close when you touch outside of the dialog, we don't want that please add this code in line 104
`datePickerDialog.setCanceledOnTouchOutside(false);`

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateParseException.java
```diff
@@ -0,0 +1,16 @@
+package com.hyperwallet.android.ui.view.widget;
+
+public class DateParseException extends Exception {
```
peter-joseph: don't make this public please we don't want to introduce new Exception handling routine for other UI's

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateDialogFragment.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Calendar;
+
+public class DateDialogFragment extends DialogFragment {
+
+    public static final String TAG = DateDialogFragment.class.getName();
+    private static final String ARGUMENT_DATE = "ARGUMENT_DATE";
+    private static final String ARGUMENT_FIELD_NAME = "ARGUMENT_FIELD_NAME";
+    private OnSelectedDateCallback mOnSelectedDateCallback;
+    private DateUtil mDateUtil;
```
peter-joseph: why not initialize date util as final member?

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateDialogFragment.java
```diff
@@ -0,0 +1,118 @@
...
+                        if (which == DialogInterface.BUTTON_NEGATIVE) {
+                            mOnSelectedDateCallback.setSelectedDateField(fieldName, null);
+                        }
+                    }
+                });
+
+        return datePickerDialog;
+    }
+
+    private DateUtil getDateUtil() {
```
peter-joseph: if you initialized this as final member then you wont need this routine

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateParseException.java
azakrevska-epam: Done

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateDialogFragment.java
azakrevska-epam: @peter-joseph Done

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateParseException.java
```diff
@@ -0,0 +1,16 @@
+package com.hyperwallet.android.ui.view.widget;
+
+class DateParseException extends Exception {
```
fmattos-hw: We actually don't need this class because of the way we handle the exception in WidgetDateDialogFragment$onCreateDialog. We can let the parse exception bubble up

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateUtil.java
```diff
@@ -0,0 +1,118 @@
...
+     * DateFormat}  or (dd MMMM yyyy)
+     *
+     * @param serverDate Date from server
+     * @return String date in the format by BestDateTimePattern or dd MMMM yyyy otherwise
+     * @throws DateParseException unable to convert server date to widget format
+     */
+    @NonNull
+    String convertDateFromServerToWidgetFormat(@Nullable final String serverDate) throws DateParseException {
+        try {
+            if (isServerDateValid(serverDate)) {
```
fmattos-hw: we can assume the server data should always be correct(if present - meaning no parse exception handling). We can remove that method and its usages because the way we handle the DateParseException is the same we would handle any other exception. So in that case, no reason to catch those exceptions, throw a new one and handle that specific exception as a generic exception

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateUtil.java
```diff
@@ -0,0 +1,118 @@
...
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.Locale;
+
+/**
+ * Class is used for manage and convert date {@link DateWidget}
+ */
+public final class DateUtil {
```
fmattos-hw: In this class we should not handle any parsing exception. if the data comes from the server, it should be correct. If we do run into a parse exception, there's no difference between catching it and throwing DateParseException because both would be handled the same way. So in isServerDateValid method, we just validate if the date is actually present (not empty) if it is not empty, do not handle parse exception

ui/src/test/java/com/hyperwallet/android/ui/view/widget/DateUtilTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+
+import java.util.Arrays;
+import java.util.Calendar;
+import java.util.Collection;
+import java.util.List;
+
+@RunWith(RobolectricTestRunner.class)
+public class DateUtilTest {
```
fmattos-hw: test names are not following the naming convention, testMethodName_UseCase
also it would be better to use the ExpectedException if possible

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateUtil.java
```diff
@@ -0,0 +1,118 @@
...
+     */
+    @NonNull
+    String convertDateFromServerToWidgetFormat(@Nullable final String serverDate) throws DateParseException {
+        try {
+            if (isServerDateValid(serverDate)) {
+                final Calendar calendar = Calendar.getInstance();
+                calendar.setTime(mServerDateFormat.parse(serverDate));
+                return mWidgetDateFormat.format(calendar.getTime());
+            }
+        } catch (NumberFormatException | ParseException | IndexOutOfBoundsException e) {
```
fmattos-hw: what would be the use case where we could have a IndexOutOfBoundsException or NumberFormatException?
unless there's not a valid case, we can just throw the ParseException as part of the method signature

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateUtil.java
```diff
@@ -0,0 +1,118 @@
...
+        } catch (ParseException e) {
+            throw new DateParseException("Unable to convert server date to calendar format", e.getCause());
+        }
+        return calendar;
+    }
+
+    private boolean isServerDateValid(@Nullable final String serverDate) throws DateParseException {
+        Date date = null;
+        if (!TextUtils.isEmpty(serverDate)) {
+            SimpleDateFormat formatter = new SimpleDateFormat(SERVER_DATE_PATTERN, Locale.getDefault());
```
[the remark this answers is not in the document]
azakrevska-epam: @fmattos-hw Thank you, done

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateUtil.java
azakrevska-epam: Done. I added throws ParseException

ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateUtil.java
azakrevska-epam: Done

ui/src/test/java/com/hyperwallet/android/ui/view/widget/DateUtilTest.java
azakrevska-epam: @fmattos-hw Thank you. Done

PR #31, opened 2019-05-30
HW-52584 receipt list

peter-joseph:


common/src/main/java/com/hyperwallet/android/common/util/DateUtility.java
```diff
@@ -0,0 +1,92 @@
...
+package com.hyperwallet.android.common.util;
+
+import androidx.annotation.NonNull;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Locale;
+
+public final class DateUtility {
```
fmattos-hw: Let's stick with DateUtils to keep it consistent with what we have so far
Also class should have a private default constructor

common/src/test/java/com/hyperwallet/android/common/util/DateUtilityTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import org.junit.Test;
+
+import java.util.Date;
+
+public class DateUtilityTest {
```
fmattos-hw: just being a little picky here... but use returns for all test names instead of return, ie. testToDateFormat_returnExpectedStringFormat would change to testToDateFormat_returnsExpectedStringFormat

receipt/build.gradle
```diff
@@ -2,6 +2,16 @@ apply from: "$rootProject.projectDir/android-library.gradle"
 
 dependencies {
     api project(":common")
+
+    implementation "com.google.android.material:material:1.0.0"
```
fmattos-hw: are these implementation dependencies temporary? If not I think we should add the version as variable. If I'm not mistaken @azakrevska-epam did that in another pr

receipt/src/main/java/com/hyperwallet/android/receipt/repository/ReceiptDataSource.java
```diff
@@ -0,0 +1,160 @@
...
+import com.hyperwallet.android.model.paging.HyperwalletPageList;
+import com.hyperwallet.android.model.transfermethod.HyperwalletTransferMethod;
+import com.hyperwallet.android.model.transfermethod.HyperwalletTransferMethodPagination;
+
+public class ReceiptDataSource extends PageKeyedDataSource<Integer, HyperwalletTransferMethod> {
+
+    private final HyperwalletTransferMethodPagination mReceiptPagination;
+    private final MutableLiveData<Boolean> mIsFetchingData = new MutableLiveData<>();
+    private final MutableLiveData<HyperwalletErrors> mErrors = new MutableLiveData<>();
+    private final Handler mHandler = new Handler();
```
fmattos-hw: Handler is not necessary, it can be null when we use view model

receipt/src/main/java/com/hyperwallet/android/receipt/repository/ReceiptDataSource.java
```diff
@@ -0,0 +1,160 @@
...
+    private LoadInitialCallback<Integer, HyperwalletTransferMethod> mLoadInitialCallback;
+    private LoadInitialParams<Integer> mLoadInitialParams;
+    private LoadCallback<Integer, HyperwalletTransferMethod> mLoadAfterCallback;
+    private LoadParams<Integer> mLoadAfterParams;
+
+    ReceiptDataSource() {
+        super();
+        mReceiptPagination = new HyperwalletTransferMethodPagination();
+        mReceiptPagination.setStatus(ACTIVATED);
+        mReceiptPagination.setSortBy(DESCENDANT_CREATE_ON);
```
fmattos-hw: don't forget to use the builder when it's merged

receipt/src/main/java/com/hyperwallet/android/receipt/repository/ReceiptRepositoryImpl.java
```diff
@@ -0,0 +1,74 @@
...
+    ReceiptRepositoryImpl() {
+        mDataSourceFactory = new ReceiptDataSourceFactory();
+        mReceiptDataSourceLiveData = mDataSourceFactory.getReceiptDataSource();
+    }
+
+    @Override
+    public LiveData<PagedList<Receipt>> loadReceipts() {
+        if (mReceiptsLiveData == null) {
+            PagedList.Config config = new PagedList.Config.Builder()
+                    .setPageSize(10)
```
azakrevska-epam: @peter-joseph Maybe it is better to extract pageSize and LoadSizeHint magic numbers to static final variables? i.e.
private static final int PAGE_SIZE = 10;
private static final int INITIAL_LOAD_SIZE = 20;

receipt/src/main/java/com/hyperwallet/android/receipt/repository/ReceiptDataSourceFactory.java
```diff
@@ -0,0 +1,46 @@
...
+public class ReceiptDataSourceFactory extends DataSource.Factory {
+
+    private final MutableLiveData<ReceiptDataSource> mDataSourceMutableLiveData;
+    private final ReceiptDataSource mReceiptDataSource;
+
+    ReceiptDataSourceFactory() {
+        super();
+        mReceiptDataSource = new ReceiptDataSource();
+        mDataSourceMutableLiveData = new MutableLiveData<>();
+        mDataSourceMutableLiveData.postValue(mReceiptDataSource);
```
vshcherbyna-epam: @peter-joseph could you verify this statement, as for me they will produce the same result.

receipt/src/main/java/com/hyperwallet/android/receipt/view/ListReceiptFragment.java
```diff
@@ -0,0 +1,307 @@
...
+            String getTransactionTitle(@NonNull final String receiptType, final int numberOfCharsAlreadyUsed,
+                    @NonNull final Context context) {
+                String showTitle = context.getResources().getString(R.string.unknown_type);
+                int resourceId = context.getResources().getIdentifier(receiptType.toLowerCase(Locale.ROOT), "string",
+                        context.getPackageName());
+                if (resourceId != 0) {
+                    showTitle = context.getResources().getString(resourceId);
+                }
+
+                if (!context.getResources().getBoolean(R.bool.isLandscape)
```
azakrevska-epam: @peter-joseph Maybe we can use "ellipsize" ? (https://developer.android.com/reference/android/widget/TextView.html#attr_android:ellipsize) and why MAX_CHARACTERS_FIRST_LINE is 38?

receipt/src/main/res/layout/item_receipt.xml
```diff
@@ -0,0 +1,70 @@
...
+
+    <TextView android:id="@+id/transaction_title"
+              android:layout_width="wrap_content"
+              android:layout_height="wrap_content"
+              android:layout_marginStart="@dimen/grid_margin_left"
+              android:layout_marginTop="@dimen/item_text_top_bottom_margin"
+              android:text="@string/text_placeholder"
+              android:textAppearance="@style/TextAppearance.MaterialComponents.Subtitle1"
+              app:layout_constraintStart_toEndOf="@+id/transaction_type_icon"
+              app:layout_constraintTop_toTopOf="parent"/>
```
azakrevska-epam: @peter-joseph  We can use tools:text - for design and can remove android:text 
    <TextView android:id="@+id/transaction_title"
              android:layout_width="0dp"
              android:layout_height="wrap_content"
              android:layout_marginStart="@dimen/grid_margin_left"
              android:layout_marginTop="@dimen/item_text_top_bottom_margin"
              android:layout_marginEnd="10dp"
              android:ellipsize="end"
              android:maxLines="1"
              android:textAppearance="@style/TextAppearance.MaterialComponents.Subtitle1"
              app:layout_constraintEnd_toStartOf="@+id/transaction_amount"
              app:layout_constraintStart_toEndOf="@+id/transaction_type_icon"
              app:layout_constraintTop_toTopOf="parent"
              tools:text="Some quite long-long-long-long-long-long-long-long-long-long-string"/>

receipt/src/main/java/com/hyperwallet/android/receipt/view/ListReceiptFragment.java
```diff
@@ -0,0 +1,307 @@
...
+                if (isDataViewType(previous, current)) {
+                    return DATA_VIEW_TYPE;
+                }
+            }
+            return HEADER_VIEW_TYPE;
+        }
+
+        boolean isDataViewType(@NonNull final Receipt previous, @NonNull final Receipt current) {
+            Calendar prev = Calendar.getInstance();
+            prev.setTime(DateUtils.fromDateTimeString(previous.getCreatedOn()));
```
vshcherbyna-epam: @peter-joseph could we avoid parsing String in a Adapter, it's quite time-consuming operation. 
I would prefer holding Date in a Receipt. This will save time twice.

receipt/src/main/java/com/hyperwallet/android/receipt/view/ListReceiptFragment.java
```diff
@@ -0,0 +1,307 @@
...
+        }
+
+        @Override
+        public boolean areContentsTheSame(@NonNull Receipt oldItem, @NonNull Receipt newItem) {
+            return oldItem.hashCode() == newItem.hashCode()
+                    && Objects.equals(oldItem, newItem);
+        }
+    }
+
+    private static class ListReceiptAdapter
```
vshcherbyna-epam: probably it would be better to split adapter in a separate class.

receipt/src/main/java/com/hyperwallet/android/receipt/view/ListReceiptFragment.java
```diff
@@ -0,0 +1,307 @@
...
+                Receipt previous = getItem(position - 1);
+                Receipt current = getItem(position);
+                if (isDataViewType(previous, current)) {
+                    return DATA_VIEW_TYPE;
+                }
+            }
+            return HEADER_VIEW_TYPE;
+        }
+
+        boolean isDataViewType(@NonNull final Receipt previous, @NonNull final Receipt current) {
```
azakrevska-epam: @peter-joseph Maybe we shouldn't do it in the adapter?

receipt/src/main/java/com/hyperwallet/android/receipt/view/ListReceiptFragment.java
```diff
@@ -0,0 +1,307 @@
...
+                super(item);
+                mTransactionAmount = item.findViewById(R.id.transaction_amount);
+                mTransactionCurrency = item.findViewById(R.id.transaction_currency);
+                mTransactionDate = item.findViewById(R.id.transaction_date);
+                mTransactionTitle = item.findViewById(R.id.transaction_title);
+                mTransactionTypeIcon = item.findViewById(R.id.transaction_type_icon);
+            }
+
+            void bind(@NonNull final Receipt receipt) {
+                if (CREDIT.equals(receipt.getEntry())) {
```
azakrevska-epam: @peter-joseph Maybe it can be done via styles or just two layouts?

receipt/src/main/java/com/hyperwallet/android/receipt/view/ListReceiptFragment.java
peter-joseph: Yep I like that approach actually it will save us time, what do you think @azakrevska-epam

receipt/src/main/java/com/hyperwallet/android/receipt/view/ListReceiptFragment.java
peter-joseph: @vshcherbyna-epam what do you mean by this? can you expand your wisdom about your comment?

receipt/src/main/java/com/hyperwallet/android/receipt/view/ListReceiptFragment.java
peter-joseph: Its ok to do this here we are not saving it, it will be GC'd after we don't use it right away, but we cannot save date instance to Receipt as well since createdOn in Receipt object is GMT, and in presentation layer we need to change it on whatever locale is the mobile user using so no.. this is good enough

receipt/src/main/java/com/hyperwallet/android/receipt/view/ListReceiptFragment.java
peter-joseph: In second thought, we cannot save date instance to Receipt as well since createdOn in Receipt object is GMT, and in presentation layer we need to change it on whatever locale is the mobile user is using.

This code is as good as it can be

receipt/src/main/java/com/hyperwallet/android/receipt/repository/ReceiptDataSourceFactory.java
peter-joseph: I think its safe @vshcherbyna-epam 
Please have a look on doc 
https://developer.android.com/reference/android/arch/lifecycle/LiveData.html#postValue(T)

receipt/src/main/java/com/hyperwallet/android/receipt/view/ListReceiptFragment.java
vshcherbyna-epam: I was just looking onto moving the Adapter to separate file and keep Fragment a little bit neat.

receipt/src/main/java/com/hyperwallet/android/receipt/repository/ReceiptDataSourceFactory.java
vshcherbyna-epam: Thanks, I saw docs, but could we leave either postValue or setValue?

common/src/main/java/com/hyperwallet/android/common/util/DateUtils.java
```diff
@@ -0,0 +1,95 @@
...
+package com.hyperwallet.android.common.util;
+
+import androidx.annotation.NonNull;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Locale;
+
+public final class DateUtils {
```
fmattos-hw: I understand the reason we are instantiating the date formatters in each method call, but the chances of someone switching the locale/timezone while using the app are pretty small. I'd say we keep those date format as static final members of the class. Thoughts @peter-joseph @azakrevska-epam @vshcherbyna-epam ?

receipt/src/main/java/com/hyperwallet/android/receipt/view/ListReceiptFragment.java
```diff
@@ -0,0 +1,295 @@
...
+                }
+            }
+        });
+    }
+
+    void retry() {
+        mListReceiptViewModel.retryLoadReceipts();
+    }
+
+    interface OnLoadReceiptErrorCallback {
```
fmattos-hw: please don't forget to remove it based on our conversation

receipt/src/main/java/com/hyperwallet/android/receipt/view/ListReceiptFragment.java
```diff
@@ -0,0 +1,295 @@
...
+
+        void showErrorOnLoadReceipt(@NonNull final List<HyperwalletError> errors);
+    }
+
+    private static class ListReceiptItemDiffCallback extends DiffUtil.ItemCallback<Receipt> {
```
fmattos-hw: add the modifier 'final' to the method arguments when applicable

receipt/src/main/java/com/hyperwallet/android/receipt/view/ListReceiptFragment.java
```diff
@@ -0,0 +1,295 @@
...
+import com.hyperwallet.android.model.receipt.Receipt;
+import com.hyperwallet.android.receipt.R;
+import com.hyperwallet.android.receipt.viewmodel.ListReceiptViewModel;
+
+import java.util.Calendar;
+import java.util.List;
+import java.util.Locale;
+import java.util.Objects;
+
+public class ListReceiptFragment extends Fragment {
```
fmattos-hw: please add final to the method parameters when applicable

receipt/src/main/java/com/hyperwallet/android/receipt/view/ListReceiptActivity.java
```diff
@@ -0,0 +1,115 @@
...
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.appcompat.app.AppCompatActivity;
+import androidx.appcompat.widget.Toolbar;
+import androidx.fragment.app.Fragment;
+import androidx.fragment.app.FragmentManager;
+import androidx.fragment.app.FragmentTransaction;
+import androidx.lifecycle.ViewModelProviders;
+
+import com.hyperwallet.android.common.view.error.DefaultErrorDialogFragment;
```
fmattos-hw: please add final to the method parameters when applicable

common/src/test/java/com/hyperwallet/android/common/util/DateUtilsTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import org.junit.Test;
+
+import java.util.Date;
+
+public class DateUtilsTest {
```
fmattos-hw: we need a test to confirm that the gmt date returning from the server is actually being converted to the devices locale/timezone

receipt/src/main/java/com/hyperwallet/android/receipt/viewmodel/ListReceiptViewModel.java
```diff
@@ -0,0 +1,98 @@
...
+
+import com.hyperwallet.android.common.viewmodel.Event;
+import com.hyperwallet.android.model.HyperwalletError;
+import com.hyperwallet.android.model.HyperwalletErrors;
+import com.hyperwallet.android.model.receipt.Receipt;
+import com.hyperwallet.android.receipt.repository.ReceiptRepository;
+
+import java.util.List;
+
+public class ListReceiptViewModel extends ViewModel {
```
fmattos-hw: please add javadocs

receipt/src/main/res/drawable/circle_negative.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
```
fmattos-hw: What do you think of adding those dimensions into the dimensions files?

receipt/src/main/res/layout/item_receipt_with_header.xml
```diff
@@ -0,0 +1,26 @@
...
+            android:id="@+id/item_date_header_title"
+            android:layout_width="match_parent"
+            android:layout_height="match_parent"
+            android:text="@string/text_placeholder"
+            android:layout_marginStart="@dimen/grid_margin_left"
+            android:layout_marginTop="@dimen/grid_margin_top"
+            android:textAppearance="@style/TextAppearance.Hyperwallet.Body2"/>
+    </RelativeLayout>
+
+    <include layout="@layout/item_receipt"/>
```
fmattos-hw: nice 👍

receipt/src/test/java/com/hyperwallet/android/receipt/repository/ReceiptDataSourceTest.java
```diff
@@ -0,0 +1,384 @@
...
+
+        assertThat(mReceiptDataSource.getErrors().getValue(), is(notNullValue()));
+        assertThat(mReceiptDataSource.getErrors().getValue().getErrors(), Matchers.<HyperwalletError>hasSize(1));
+        assertThat(mReceiptDataSource.getErrors().getValue().getErrors().get(0).getCode(), is("TEST_CODE"));
+        assertThat(mReceiptDataSource.getErrors().getValue().getErrors().get(0).getMessage(), is("test message"));
+        assertThat(mReceiptDataSource.isFetchingData().getValue(), is(false));
+    }
+
+    @Test
+    public void testRetry_LoadInitial() {
```
fmattos-hw: it should be testRetry_loadInitial - need to change the L upper case, same with testRetry_LoadAfter

receipt/src/test/java/com/hyperwallet/android/receipt/repository/ReceiptDataSourceTest.java
```diff
@@ -0,0 +1,384 @@
...
+        assertThat(receipts.get(3).getJournalId(), is("51660675"));
+        assertThat(receipts.get(3).getType(), is(PAYMENT));
+        assertThat(receipts.get(3).getEntry(), is(CREDIT));
+        assertThat(receipts.get(3).getSourceToken(), is("act-b1f6dc28-e534-45f4-a661-3523f051f77a"));
+        assertThat(receipts.get(3).getDestinationToken(), is("usr-b4e8ec34-52d8-4a81-9566-bdde1bd745b6"));
+        assertThat(receipts.get(3).getAmount(), is("13.00"));
+        assertThat(receipts.get(3).getFee(), is("0.00"));
+        assertThat(receipts.get(3).getCurrency(), is("USD"));
+        assertThat(receipts.get(3).getDetails(), is(notNullValue()));
+        assertThat(receipts.get(3).getDetails().getPayeeName(), is("kevin Puckett"));
```
fmattos-hw: just being a little picky here, use upper case in Kevin. there are more cases like that too

ui/src/main/java/com/hyperwallet/android/ui/repository/TransferMethodRepositoryImpl.java
```diff
@@ -66,7 +68,11 @@ public void createTransferMethod(@NonNull final HyperwalletTransferMethod transf
 
     @Override
     public void loadTransferMethods(@NonNull final LoadTransferMethodListCallback callback) {
-        getHyperwallet().listTransferMethods(null,
+
+        HyperwalletTransferMethodQueryParam queryParam = new HyperwalletTransferMethodQueryParam.Builder()
```
fmattos-hw: 🥇

common/src/main/java/com/hyperwallet/android/common/util/DateUtils.java
```diff
@@ -0,0 +1,95 @@
...
+            throw new IllegalArgumentException("An exception occurred when attempting to parse " +
+                    "the date " + dateString, e);
+        }
+    }
+
+
+    /**
+     * date format: yyyy-MM-dd'T'HH:mm:ss
+     */
+    public static Date fromDateTimeString(@NonNull final String dateString, @NonNull final String format) {
```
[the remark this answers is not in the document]
peter-joseph: reduced scope and made it visible for testing

common/src/main/java/com/hyperwallet/android/common/viewmodel/Event.java
```diff
@@ -0,0 +1,54 @@
...
+        mIsContentConsumed = true;
+        return content;
+    }
+
+    public boolean isContentConsumed() {
+        return mIsContentConsumed;
+    }
+
+    /**
+     * retrieve perceived content based on iff content has not been retrieved
```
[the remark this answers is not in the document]
peter-joseph: short hand of if and only if :), will change to laymens term

common/src/main/java/com/hyperwallet/android/common/util/DateUtils.java
peter-joseph: making it static will only make it more susceptible for memory leak, for me will leave it as it is since if call stack of that function is popped that means all its variable are also gc'd

common/src/test/java/com/hyperwallet/android/common/util/DateUtilsTest.java
peter-joseph: hmm formatting of string to different locale is not supported in unit test, that means it will only work on automation test, probably we can add that to automation test

receipt/src/main/java/com/hyperwallet/android/receipt/repository/ReceiptRepositoryImpl.java
```diff
@@ -0,0 +1,92 @@
...
+
+import com.hyperwallet.android.model.HyperwalletErrors;
+import com.hyperwallet.android.model.receipt.Receipt;
+
+/**
+ * {@link ReceiptRepository} implementation
+ */
+public class ReceiptRepositoryImpl implements ReceiptRepository {
+
+    private static final int PAGE_SIZE = 10;
```
skoong: Should this be 20 as well?

peter-joseph: Sorry no default size is still, 10, prefetch is 20

common/src/test/java/com/hyperwallet/android/common/util/DateUtilsTest.java
peter-joseph: found a way, we might be mis understanding androids formatter

receipt/src/test/java/com/hyperwallet/android/receipt/repository/ReceiptDataSourceTest.java
fmattos-hw: testRetry_LoadAfter is still with L upper case

PR #40, opened 2019-06-12
HW-53384:  Update UI group id

vshcherbyna-epam:


peter-joseph:
In the receipts module I see that the test sources are not full migrated
Screen Shot 2019-06-13 at 10 03 43 AM

PR #45, opened 2019-06-18
HW-53765. renamed *Util to *Utils

azakrevska-epam:


ui/src/main/java/com/hyperwallet/android/ui/view/widget/DateUtils.java
```diff
@@ -31,15 +31,15 @@
 /**
  * Class is used for manage and convert date {@link DateWidget}
  */
-public final class DateUtil {
+public final class DateUtils {
```
peter-joseph: Sorry I forgot we have this, `DateUtil` class in ui module, @azakrevska-epam @fmattos-hw there's also a `DateUtils` class in `com.hyperwallet.android.ui.common.util.DateUtils` does it make sense to merge them into one?

fmattos-hw: I think the class com.hyperwallet.android.ui.view.widget.DateUtils should not exist and all methods should be moved to the DateWidget class

PR #49, opened 2019-06-19
HW-53039 receipt view ppc

peter-joseph:


receipt/src/main/java/com/hyperwallet/android/ui/receipt/view/ListReceiptFragment.java
```diff
@@ -199,25 +199,25 @@ public void onBindViewHolder(@NonNull final ReceiptViewHolder holder, final int
 
         class ReceiptViewHolder extends RecyclerView.ViewHolder {
 
-            private ListReceiptViewModel mListReceiptViewModel;
+            private ReceiptViewModel mListUserReceiptViewModel;
             private View mView;
 
-            ReceiptViewHolder(@NonNull final ListReceiptViewModel receiptViewModel,
+            ReceiptViewHolder(@NonNull final ReceiptViewModel receiptViewModel,
```
azakrevska-epam: @peter-joseph Maybe I missed something, but why do we implement interaction with viewModel directly? Can we use some interface for this? for example:
private OnReceiptClickListener ...
onReceiptClickListener.clicked(receipt) and implement it in the Fragment?

peter-joseph: yeah but we don't need to make this as a pattern that is why the code is inclusive, we only want to click on data part not the header part of the recycle view item.

PR #53, opened 2019-06-25
Removing unused exceptions

skoong:


skoong:
Addressed in https://github.com/hyperwallet/hyperwallet-android-ui-sdk/pull/49

PR #70, opened 2019-07-22
HW-54790. Transfer UI module

azakrevska-epam:


azakrevska-epam:
Closed due to parent branch

PR #71, opened 2019-07-22
Single jacoco

azakrevska-epam:


fmattos-hw:
@azakrevska-epam I really like the idea. Please see if userrepository module also need changes and create a PR to development branch instead.
Thanks

azakrevska-epam:
Due to changed parent branch

PR #75, opened 2019-07-23
Updated jacoco settings

azakrevska-epam:


azakrevska-epam:
Unable to push changes. see #78

PR #63, opened 2019-07-15
HW-54255 create unit test on viewmodels

peter-joseph:

PR #77, opened 2019-07-23
HW-54255 unit test for view models

peter-joseph:
From https://github.com/hyperwallet/hyperwallet-android-ui-sdk/pull/63


receiptui/src/test/java/com/hyperwallet/android/ui/receipt/rule/HyperwalletExternalResourceManager.java
```diff
@@ -1,4 +1,4 @@
-package com.hyperwallet.android.ui.transfermethod.rule;
+package com.hyperwallet.android.ui.receipt.rule;
```
azakrevska-epam: com.hyperwallet.android.ui.*receipt*.rule inside of   .../hyperwallet/android/ui/*transfermethod*/rule ?

PR #79, opened 2019-07-24
HW-54789, 54794. Transfer Destination ViewModel

azakrevska-epam:


transferui/src/main/java/com/hyperwallet/android/ui/transfer/viewmodel/ListTransferDestinationViewModel.java
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * @see ViewModel#onCleared()
+     */
+    @Override
+    protected void onCleared() {
+        super.onCleared();
+        mTransferMethodRepository = null;
+    }
+
```
azakrevska-epam: @fmattos-hw Should we test ListTransferDestinationViewModelFactory? 
If yes - what kind of class should I pass to the create() to get IllegalArgumentException?

peter-joseph: we can do same stuff in create transfer viewmodel test
in commit link -> https://github.com/hyperwallet/hyperwallet-android-ui-sdk/pull/83/commits/57a68a76c4e71bb3641e8d94720313b15fda7a4b

Have a look on test class `CreateTransferViewModelTest` methods `testCreateTransferViewModelFactory_createCreateTransferViewModelUnSuccessful and testCreateTransferViewModelFactory_createCreateTransferViewModelSuccessful`

vshcherbyna-epam: added

PR #80, opened 2019-07-24
Moving test util and rules into separate module

skoong:
Screen Shot 2019-07-24 at 10 50 55 AM


testutils/config/jacoco-settings.gradle
```diff
@@ -0,0 +1,79 @@
+apply plugin: 'jacoco'
```
vshcherbyna-epam: Shyang, wouldn't you use a single jacoco file for this purpose? Like there https://github.com/hyperwallet/hyperwallet-android-ui-sdk/pull/78

skoong: I was just following whatever was already there.  I can update it after we merge that one.

PR #83, opened 2019-07-25
HW-54672 create transfer ui

peter-joseph:


transferui/src/main/java/com/hyperwallet/android/ui/transfer/viewmodel/CreateTransferViewModel.java
```diff
@@ -0,0 +1,338 @@
...
+        createLoadErrorObserver();
+        loadTransferSource();
+        createQuoteErrorObserver();
+    }
+
+    public LiveData<Boolean> isTransferAllAvailableFunds() {
+        return mTransferAvailableFunds;
+    }
+
+    public void setTransferAllAvailableFunds(Boolean transferAll) {
```
fmattos-hw: @nonnull/final ?

transferui/src/main/java/com/hyperwallet/android/ui/transfer/viewmodel/CreateTransferViewModel.java
```diff
@@ -0,0 +1,338 @@
...
+            }
+        };
+        mQuoteErrors.observeForever(mQuoteEventObserver);
+    }
+
+    private void loadTransferSource() {
+        mUserRepository.loadUser(new UserRepository.LoadUserCallback() {
+            @Override
+            public void onUserLoaded(@NonNull HyperwalletUser user) {
+                loadTransferDestination(user.getToken());
```
fmattos-hw: heads up, we will need a member variable to hold the user token value - (error handling/retry use cases) - I will leave it up to you to handle it now or in the future sprints

transferui/src/main/java/com/hyperwallet/android/ui/transfer/view/CreateTransferFragment.java
```diff
@@ -0,0 +1,379 @@
...
+                }
+            }
+
+            @Override
+            public void afterTextChanged(Editable s) {
+            }
+        });
+    }
+
+    private void registerObserver() {
```
fmattos-hw: Please take a look at the documentation. This method should be split into 3, registerTransferDestinationObserver, registerAvailableFundsObserver and registerTransferAmountObserver

transferui/src/main/java/com/hyperwallet/android/ui/transfer/viewmodel/CreateTransferViewModel.java
```diff
@@ -0,0 +1,338 @@
...
+
+    // create quote error
+    private Observer<Event<HyperwalletErrors>> mQuoteEventObserver;
+    private final MutableLiveData<Event<HyperwalletErrors>> mQuoteErrorEvent = new MutableLiveData<>();
+    private final MutableLiveData<Event<HyperwalletErrors>> mQuoteErrors = new MutableLiveData<>();
+
+    /**
+     * Initialize Create Transfer View Model with designated transfer source token
+     *
+     * @param sourceToken              Source token that represents either from Wallet, Bank or other EA tokens
```
fmattos-hw: Please use user token or prepaid card token - avoid using terms like wallet/EA as they're used internally only. Also instead of HW use Hyperwallet

transferui/src/main/java/com/hyperwallet/android/ui/transfer/viewmodel/CreateTransferViewModel.java
```diff
@@ -0,0 +1,338 @@
...
+            @Override
+            public void onError(@NonNull HyperwalletErrors errors) {
+                mIsLoading.postValue(Boolean.FALSE);
+                mLoadErrors.postValue(new Event<>(errors));
+            }
+        });
+    }
+
+    private void loadTransferDestination(@NonNull final String sourceToken) {
+        mTransferMethodRepository.loadLatestTransferMethod(new TransferMethodRepository.LoadTransferMethodCallback() {
```
fmattos-hw: I can't find its implementation - but we should set the limit size to 1 as part of the query param

transferui/src/main/java/com/hyperwallet/android/ui/transfer/viewmodel/CreateTransferViewModel.java
```diff
@@ -0,0 +1,338 @@
...
+import com.hyperwallet.android.ui.user.repository.UserRepository;
+
+import java.util.UUID;
+
+/**
+ * Create Transfer View Model
+ */
+public class CreateTransferViewModel extends ViewModel {
+
+    private static final String CLIENT_IDENTIFICATION_PREFIX = "HW-MBLA-";
```
fmattos-hw: let's use HW-ANDROID as prefix

transferui/src/main/res/layout/fragment_create_transfer.xml
```diff
@@ -0,0 +1,242 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout
```
fmattos-hw: I'm seeing a few hardcoded dp values. Add them to dimens.xml if it makes sense

transferui/src/main/res/layout/fragment_create_transfer.xml
```diff
@@ -0,0 +1,242 @@
...
+                    android:id="@+id/transfer_amount_currency"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:layout_alignTop="@+id/transfer_amount"
+                    android:layout_alignEnd="@+id/transfer_amount"
+                    android:layout_alignBottom="@+id/transfer_amount"
+                    android:layout_marginTop="13dp"
+                    android:layout_marginEnd="4dp"
+                    android:layout_marginBottom="0dp"
+                    android:text="@string/transfer_currency_placeholder"/>
```
fmattos-hw: I think it's better to use tools:text="@string/transfer_currency_placeholder"

transferui/src/test/java/com/hyperwallet/android/ui/transfer/viewmodel/CreateTransferViewModelTest.java
```diff
@@ -0,0 +1,298 @@
...
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+import org.mockito.stubbing.Answer;
+import org.robolectric.Robolectric;
+import org.robolectric.RobolectricTestRunner;
+
+import java.util.Date;
+
+@RunWith(RobolectricTestRunner.class)
+public class CreateTransferViewModelTest {
```
fmattos-hw: Please add a test case to verify the default values (testCreateTransferViewModel_verifyDefaultValues maybe...) in both constructors. I've seen some comments about the default values (you can keep those), but I think a test case for the default values is important

transferui/src/main/java/com/hyperwallet/android/ui/transfer/viewmodel/CreateTransferViewModel.java
```diff
@@ -0,0 +1,338 @@
...
+    private final TransferMethodRepository mTransferMethodRepository;
+    private final UserRepository mUserRepository;
+
+    private final MutableLiveData<HyperwalletTransferMethod> mTransferDestination = new MutableLiveData<>();
+    private final MutableLiveData<Boolean> mTransferAvailableFunds = new MutableLiveData<>();
+    private final MutableLiveData<Boolean> mIsLoading = new MutableLiveData<>();
+    private final MutableLiveData<Boolean> mIsCreateQuoteLoading = new MutableLiveData<>();
+    private final MutableLiveData<Transfer> mQuoteAvailableFunds = new MutableLiveData<>();
+    private final MutableLiveData<Transfer> mQuoteTransfer = new MutableLiveData<>();
+    private final MutableLiveData<String> mTransferAmount = new MutableLiveData<>();
```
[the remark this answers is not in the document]
peter-joseph: Since we are using LiveData to manage data lifecycle, might as well use this in data form fields, which in this turn give us the ability to delegate android lifecycle destruction/construction of data to LiveData framework (we don't need to manage phone rotation, app back stack and app sleep mode), this also in turn gives us the advantage of easy maintainability since there's only 1 single truth of data source which is our ViewModel (No need to look over the lifecycle listener where the data changed, we always know that its from view model + component)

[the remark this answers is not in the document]
fmattos-hw: 👍

PR #85, opened 2019-07-31
HW-53671 add select transfer destination

peter-joseph:


commonui/src/main/java/com/hyperwallet/android/ui/common/view/OnItemClickListener.java
```diff
@@ -0,0 +1,25 @@
+/*
```
fmattos-hw: I don't think we should use this approach. At first, it seems like a good idea but when I look deeper at it, I have the following thoughts:
* The class implementing this interface is the upper class, the adapter, which is responsible for inflating the view holders and managing the data that is sent to the recycler view, AND handling the clicks on each view holder, while in fact, it should be the view holder itself responsible for handling its events. I understand that the view holder is invoking/delegating the event handling to the upper class which I think isn't the ideal approach. In my opinion, the view holder should handle the click and not delegate it to the upper layer. 
* Since the onBindViewHolder already gives us the position and an easy way to get the destination (transfer method object), it makes it more appealing to use a listener there and let each view holder being an independent "object" since each one of them will have its own listener as opposed to all view holder sharing the same listener (the adapter)
We can discuss this offline if needed

transferui/src/main/java/com/hyperwallet/android/ui/transfer/view/CreateTransferFragment.java
```diff
@@ -129,19 +136,22 @@ public void onClick(View v) {
...
         // transfer destination
         mTransferDestination = view.findViewById(R.id.transfer_destination);
-        mTransferDestination.setOnClickListener(new View.OnClickListener() {
+        mTransferDestination.setOnClickListener(new OneClickListener() {
             @Override
-            public void onClick(View v) {
-                //TODO temporary action for transfer destination clicked; it should open up transfer method list UI
-                Snackbar.make(mTransferDestination, "LIST Transfer clicked", Snackbar.LENGTH_SHORT).show();
+            public void onOneClick(View v) {
+                HyperwalletTransferMethod activeDestination =
```
fmattos-hw: The only reason we need the current destination is for the token comparison in the adapter, isn't it? If that's the case, I'd rather to pass the token and not the whole object, then we can avoid the whole parcelable serialization of the whole object

transferui/src/main/java/com/hyperwallet/android/ui/transfer/view/ListTransferDestinationFragment.java
```diff
@@ -0,0 +1,287 @@
...
+     * Please don't use this constructor this is reserved for Android Core Framework
+     *
+     * @see #newInstance(HyperwalletTransferMethod)
+     */
+    public ListTransferDestinationFragment() {
+    }
+
+    static ListTransferDestinationFragment newInstance(@NonNull final HyperwalletTransferMethod transferMethod) {
+        Bundle bundle = new Bundle();
+        bundle.putParcelable(ARGUMENT_SELECTED_TRANSFER, transferMethod);
```
fmattos-hw: I think it would be better to set it to the view model, what do you think? Then we wouldn't need to implement onViewStateRestored

transferui/src/main/java/com/hyperwallet/android/ui/transfer/view/ListTransferDestinationFragment.java
```diff
@@ -0,0 +1,287 @@
...
+                if (loading) {
+                    mProgressBar.setVisibility(View.VISIBLE);
+                } else {
+                    mProgressBar.setVisibility(View.GONE);
+                }
+            }
+        });
+    }
+
+    void hideSoftKey(@NonNull View focusedView) {
```
fmattos-hw: can't we make it private and also onClose and onView too?

transferui/src/main/java/com/hyperwallet/android/ui/transfer/view/ListTransferDestinationFragment.java
```diff
@@ -0,0 +1,287 @@
...
+        public void onViewRecycled(@NonNull TransferDestinationViewHolder holder) {
+            holder.recycle();
+        }
+
+        @Override
+        public void onItemClick(int position) {
+            mViewModel.selectedTransferDestination(mTransferDestinations.get(position));
+        }
+
+        void replaceData(List<HyperwalletTransferMethod> destinations) {
```
fmattos-hw: @nonnull / final if applicable

transferui/src/main/java/com/hyperwallet/android/ui/transfer/view/ListTransferDestinationFragment.java
```diff
@@ -0,0 +1,287 @@
...
+            mOnItemClickListener = itemClickListener;
+        }
+
+        @Override
+        public void onClick(View v) {
+            int position = getAdapterPosition();
+            mOnItemClickListener.onItemClick(position);
+        }
+
+        void bind(HyperwalletTransferMethod destination, boolean selected) {
```
fmattos-hw: same here @nonnull /final if applicable

transferui/src/main/java/com/hyperwallet/android/ui/transfer/view/ListTransferDestinationActivity.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    private void initFragment(Fragment fragment) {
+        FragmentManager fragmentManager = getSupportFragmentManager();
+        FragmentTransaction transaction = fragmentManager.beginTransaction();
+        transaction.add(R.id.list_destination_fragment, fragment);
+        transaction.commit();
+    }
+
+    private void registerObservers() {
```
fmattos-hw: I think it's safer to check if the content hasn't been consumed in those events...

transferui/src/main/res/layout/item_transfer_destination.xml
```diff
@@ -0,0 +1,81 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:paddingStart="@dimen/grid_padding_left"
+    android:paddingEnd="@dimen/grid_padding_right"
+    tools:background="#DADADA"
```
fmattos-hw: shouldn't this color be in the colors.xml ?

transferui/src/test/java/com/hyperwallet/android/ui/transfer/viewmodel/ListTransferDestinationViewModelTest.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Test
+    public void testInitialization_callsRepository() {
+        verify(transferMethodRepository).loadTransferMethods(
+                any(TransferMethodRepository.LoadTransferMethodListCallback.class));
+    }
+
+    @Test
+    public void testSelectTransferDestination_callRepository() {
```
fmattos-hw: please keep it consistent with the test above, use callsRepository as prefix

commonui/src/main/java/com/hyperwallet/android/ui/common/view/OnItemClickListener.java
peter-joseph: yeah thats part of Epam's code simplified it :)

transferui/src/main/java/com/hyperwallet/android/ui/transfer/view/ListTransferDestinationFragment.java
peter-joseph: I thought this is your design but simplified it

transferui/src/main/java/com/hyperwallet/android/ui/transfer/view/ListTransferDestinationActivity.java
peter-joseph: yeah but we are disposing the UI after consuming so there's no chance for it to be consume again

transferui/src/test/java/com/hyperwallet/android/ui/transfer/viewmodel/ListTransferDestinationViewModelTest.java
peter-joseph: kk migrated from EPAM's code :)

transferui/src/main/java/com/hyperwallet/android/ui/transfer/view/ListTransferDestinationFragment.java
peter-joseph: There's a bit of race condition and observer operation conflict
1) Activity creates the view model, pushed new transfer model value to ViewModel and register observer
   Fragment reference the view model and get value for selected transfer model seems to be null

2) same as 1 except that we register observer so that values are updated and set selected transfer but it will close the activity based on instructions on the observer

3) We simplified it from Transfer model to just the Token so we might not achieve this via Via view model

fmattos-hw: 👍

PR #88, opened 2019-08-07
HW-55069 Transfer ui error handling

peter-joseph:


transferui/src/test/java/com/hyperwallet/android/ui/transfer/viewmodel/CreateTransferViewModelTest.java
```diff
@@ -269,7 +269,7 @@ public void testCreateQuoteTransfer_Successful() {
         // test
```
fmattos-hw: we need to increase test coverage in this class. Please look at the build file and remove the filter 
 '**/com/hyperwallet/android/ui/transfer/viewmodel/*.*'

PR #89, opened 2019-08-08
HW-53676 Processing confirmation

peter-joseph:


transferui/src/main/res/layout/fragment_create_transfer.xml
```diff
@@ -112,19 +112,18 @@
...
 
                 <com.google.android.material.textfield.TextInputLayout
                     android:id="@+id/transfer_amount_layout"
+                    android:layout_marginStart="@dimen/grid_margin_left"
+                    android:layout_marginEnd="@dimen/grid_margin_right"
+                    android:layout_marginTop="@dimen/grid_margin_top"
                     android:layout_width="match_parent"
                     android:layout_height="wrap_content"
                     android:theme="@style/Widget.Hyperwallet.TextInputLayout"
-                    android:layout_marginStart="@dimen/edit_text_margin"
```
skoong: The alignment is a little off without these

transferui/src/main/java/com/hyperwallet/android/ui/transfer/view/ScheduleTransferFragment.java
```diff
@@ -0,0 +1,288 @@
...
+    private boolean isFeeAvailable(@Nullable final String feeAmount) {
+        if (!TextUtils.isEmpty(feeAmount)) {
+            BigDecimal fee = new BigDecimal(feeAmount.replace(CURRENCY_NUMERIC_SEPARATOR, EMPTY_STRING));
+            return fee.doubleValue() != 0;
+        }
+        return false;
+    }
+
+    // TODO this will be removed when Platform can return the total amount field
+    private String getTotalAmount(@NonNull final String amount, @NonNull final String fee) {
```
fmattos-hw: please move this to viewmodel class and create unit test

transferui/src/main/java/com/hyperwallet/android/ui/transfer/view/ScheduleTransferFragment.java
```diff
@@ -0,0 +1,288 @@
...
+        BigDecimal normalizedAmount = new BigDecimal(amount.replace(CURRENCY_NUMERIC_SEPARATOR, EMPTY_STRING));
+        BigDecimal normalizedFee = new BigDecimal(fee.replace(CURRENCY_NUMERIC_SEPARATOR, EMPTY_STRING));
+
+        BigDecimal totalAmount = normalizedAmount.add(normalizedFee);
+
+        // format in normal locale US this will change when we have formatting enabled
+        return NumberFormat.getInstance(Locale.US).format(totalAmount);
+    }
+
+    private void showNotes(@Nullable final String notes) {
```
fmattos-hw: same here

transferui/src/main/java/com/hyperwallet/android/ui/transfer/view/ScheduleTransferActivity.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    private void registerObservers() {
+        mScheduleTransferViewModel.getTransferStatusTransitionError().observe(this,
+                new Observer<Event<HyperwalletErrors>>() {
+                    @Override
+                    public void onChanged(final Event<HyperwalletErrors> event) {
+                        if (event != null && !event.isContentConsumed()) {
+                            showErrorScheduleTransfer(event.getContent().getErrors());
+                            setResult(Activity.RESULT_OK);
```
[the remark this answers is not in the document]
peter-joseph: If we are not going to dispose the activity I think we don't need to call **setResult()**, since we are not finishing the activity, there was some question closing the schedule transfer activity when there's an error, probably this is from the prototype... 

but if we are going to change the flow
*) close schedule transfer activity when there's error to go back to create transfer ui
    then we need to to do the ff:
    *.a) set result to **Activity.RESULT_CANCELED** because commit is actually cancelled after an error
    *.b) finish the activity meaning closing it and showing the create transfer UI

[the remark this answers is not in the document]
fmattos-hw: ok let's leave it as is for now

PR #94, opened 2019-08-21
HW-53672 Improve component gpu rendering

peter-joseph:


commonui/src/main/java/com/hyperwallet/android/ui/common/view/HorizontalDividerItemDecorator.java
```diff
@@ -31,17 +31,11 @@
 public class HorizontalDividerItemDecorator extends RecyclerView.ItemDecoration {
 
     protected final Drawable mHorizontalItemDivider;
-    protected final int mDefaultPadding;
+    private final int mDefaultPadding;
```
fmattos-hw: I'd just use a constant here instead

peter-joseph: ok lol

PR #103, opened 2019-10-03
HW-56154 add warning message when fx changes

fmattos-hw:
This PR contains the code changes to display a warning message to the user when the fx rates change from the "Create Transfer UI" to the "Confirm Transfer UI"


fmattos-hw:
It could have been a frame layout, relative layout as well. It wouldn't make difference in the way the UI would look like since the linear layout container is what's controlling how the elements are displayed one below another. I just tried to keep it consistent with what was already there

PR #125, opened 2019-11-19
Task/hw 55372 page tech product for error

dnip-hw:
* include missing Page Technology and Product for json string
* remove deprecated method
* using new trackError interface as per: https://github.com/hyperwallet/hyperwallet-android-insight/pull/9/files
* new params argument can be used for TransferMethod map or future considerations


commonui/src/main/java/com/hyperwallet/android/ui/common/insight/HyperwalletInsight.java
```diff
@@ -221,12 +221,19 @@ public Handler getHandler() {
     public void trackError(@NonNull final Context context, @NonNull final String pageName,
             @NonNull final String pageGroup, @NonNull final Map<String, String> errorInfoMap) {
 
+        HashMap<String, String> params = new HashMap<>(2);
+        params.put(InsightEventTag.InsightEventTagEventParams.PRODUCT,
+                errorInfoMap.get(InsightEventTag.InsightEventTagEventParams.PRODUCT));
+        params.put(InsightEventTag.InsightEventTagEventParams.PAGE_TECHNOLOGY,
+                errorInfoMap.get(InsightEventTag.InsightEventTagEventParams.PAGE_TECHNOLOGY));
```
dnip-hw: I'm not a fan of this, but not sure how else to do it. Personally, I think we should get rid of ErrorInfo object in favour of a Map so we can take advantage of loose typing.

commonui/src/main/java/com/hyperwallet/android/ui/common/insight/HyperwalletInsight.java
```diff
@@ -221,12 +231,20 @@ public Handler getHandler() {
     public void trackError(@NonNull final Context context, @NonNull final String pageName,
             @NonNull final String pageGroup, @NonNull final Map<String, String> errorInfoMap) {
 
+        HashMap<String, String> params = new HashMap<>(2);
+        for (Map.Entry<String, String> entry : errorInfoMap.entrySet()) {
+            if (!ERROR_VALUES.contains(entry.getKey())) {
+                params.put(entry.getKey(), entry.getValue());
+            }
+        }
```
dnip-hw: I'm not a fan of this, but not sure how else to do it. Personally, I think we should get rid of ErrorInfo object in favour of a Map so we can take advantage of loose typing. Since in end, they all just get thrown into "event_params"

commonui/src/main/java/com/hyperwallet/android/ui/common/insight/HyperwalletInsight.java
```diff
@@ -222,11 +232,12 @@ public void trackError(@NonNull final Context context, @NonNull final String pag
...
-                .message(errorInfoMap.get(InsightEventTag.InsightEventTagEventParams.ERROR_MESSAGE))
-                .code(errorInfoMap.get(InsightEventTag.InsightEventTagEventParams.ERROR_CODE))
-                .field(errorInfoMap.get(InsightEventTag.InsightEventTagEventParams.ERROR_FIELD_NAME))
-                .description(errorInfoMap.get(InsightEventTag.InsightEventTagEventParams.ERROR_DESCRIPTION))
+                .type(errorInfoMap.remove(InsightEventTag.InsightEventTagEventParams.ERROR_TYPE))
+                .message(errorInfoMap.remove(InsightEventTag.InsightEventTagEventParams.ERROR_MESSAGE))
+                .code(errorInfoMap.remove(InsightEventTag.InsightEventTagEventParams.ERROR_CODE))
+                .field(errorInfoMap.remove(InsightEventTag.InsightEventTagEventParams.ERROR_FIELD_NAME))
+                .description(errorInfoMap.remove(InsightEventTag.InsightEventTagEventParams.ERROR_DESCRIPTION))
+                .params(errorInfoMap)
```
dnip-hw: I'm not a fan of this.  Personally, I think we should get rid of ErrorInfo object in favour of a Map so we can take advantage of loose typing. Since in end, they all just get thrown into "event_params"

commonui/src/main/java/com/hyperwallet/android/ui/common/insight/HyperwalletInsight.java
```diff
@@ -216,21 +216,28 @@ public Handler getHandler() {
      * @param context      Context where the tracking is happening
      * @param pageName     an arbituary pageName the user is currently viewing as defined by the app
      * @param pageGroup    an arbituary pageGroup the user is currently viewing as defined by the app
+     * @param params    additional information that can be added to make the tracking event more useful
      * @param errorInfoMap contains information regarding the error
      */
     public void trackError(@NonNull final Context context, @NonNull final String pageName,
-            @NonNull final String pageGroup, @NonNull final Map<String, String> errorInfoMap) {
+            @NonNull final String pageGroup, @Nullable final Map<String, String> params,
+            @NonNull final Map<String, String> errorInfoMap) {
```
fmattos-hw: We should keep only one map parameter (params and remove errorInfoMap) in this method. 
```suggestion
public void trackError(@NonNull final Context context, @NonNull final String pageName,
            @NonNull final String pageGroup, @Nullable final Map<String, String> params) {

        final ErrorInfo errorInfo = new ErrorInfo.ErrorInfoBuilder()
                .type(params.remove(InsightEventTag.InsightEventTagEventParams.ERROR_TYPE))
                .message(params.remove(InsightEventTag.InsightEventTagEventParams.ERROR_MESSAGE))
                .code(params.remove(InsightEventTag.InsightEventTagEventParams.ERROR_CODE))
                .field(params.remove(InsightEventTag.InsightEventTagEventParams.ERROR_FIELD_NAME))
                .description(params.remove(InsightEventTag.InsightEventTagEventParams.ERROR_DESCRIPTION))
                .build();
        
        if (Insight.getInsightTracker().isInitialized()) {
            Insight.getInsightTracker().trackError(context, pageName, pageGroup, params, errorInfo);

        } else {
            mExecutor.execute(new Runnable() {
                @Override
                public void run() {
                    Hyperwallet.getDefault().getConfiguration(new HyperwalletListener<Configuration>() {
                        @Override
                        public void onSuccess(@Nullable Configuration configuration) {
                            if (configuration != null) {
                                HyperwalletInsight.getInstance().initialize(context, configuration);
                                Insight.getInsightTracker().trackError(context, pageName, pageGroup, paramsForError,
                                        errorInfo);
                            }
                        }

                        @Override
                        public void onFailure(HyperwalletException exception) {
                            // do nothing
                        }

                        @Override
                        public Handler getHandler() {
                            return null;
                        }
                    });
                }
            });
        }
    }

```

commonui/src/main/java/com/hyperwallet/android/ui/common/view/error/DefaultErrorDialogFragmentPresenter.java
```diff
@@ -21,10 +21,8 @@
 import androidx.annotation.NonNull;
```
fmattos-hw: This class and the interface being implemented can be removed. After the changes, this class is no longer necessary

commonui/src/main/java/com/hyperwallet/android/ui/common/util/ErrorUtils.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.List;
+
+/**
+ * Common HW-SDK UI Date Utility class, that will assist on safe presentation of date whatever the mobile device setting
+ * is set Locale, Timezone and etc... that dictates how that dates are being presented
+ *
+ * Moreover all date string to {@link Date} object conversion is automatically converted from
+ * GMT date string from API to locale Date set by the phone
+ */
+public final class ErrorUtils {
```
fmattos-hw: please add private default constructor

commonui/src/main/java/com/hyperwallet/android/ui/common/util/ErrorUtils.java
```diff
@@ -0,0 +1,69 @@
...
+import androidx.annotation.NonNull;
+
+import com.hyperwallet.android.model.HyperwalletError;
+
+import java.util.Date;
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Common HW-SDK UI Date Utility class, that will assist on safe presentation of date whatever the mobile device setting
```
fmattos-hw: javadoc copied from another class...

commonui/src/main/java/com/hyperwallet/android/ui/common/util/ErrorUtils.java
```diff
@@ -0,0 +1,69 @@
...
+ * is set Locale, Timezone and etc... that dictates how that dates are being presented
+ *
+ * Moreover all date string to {@link Date} object conversion is automatically converted from
+ * GMT date string from API to locale Date set by the phone
+ */
+public final class ErrorUtils {
+
+    public static String getMessage(@NonNull List<HyperwalletError> errors, @NonNull Resources resources) {
```
fmattos-hw: please add javadocs

commonui/src/main/java/com/hyperwallet/android/ui/common/insight/HyperwalletInsight.java
manikandan-altmtrk: I think, here we are confusing more. We can send as ErrorInfo from where the event occurs and also here we can get the same as per confluence. Then we can process the same EventProcessor.java in inside project. Those params also describes the error in detail. So, i think params also part of error info. 
Current scenario, We need to send the **"null"** value HashMap as a argument if params are not required.
Other options, I think recommended way to call a method with different number of argument by using Method Overloading.
    public void trackError(@NonNull final Context context, @NonNull final String pageName,
            @NonNull final String pageGroup, @NonNull final Map<String, String> errorInfoMap) {
trackError(context, pageName,pageGroup, null, errorInfoMap);// instead of sending everywhere as a null
}
    public void trackError(@NonNull final Context context, @NonNull final String pageName,
            @NonNull final String pageGroup, @Nullable final Map<String, String> params,
            @NonNull final Map<String, String> errorInfoMap) {
//Actual code
}
Screenshot 2019-11-20 at 8 23 26 PM
Screenshot 2019-11-20 at 8 43 14 PM
Screenshot 2019-11-20 at 8 57 26 PM

commonui/src/main/java/com/hyperwallet/android/ui/common/view/ActivityUtils.java
```diff
@@ -64,10 +65,15 @@ public static void showError(@NonNull final FragmentActivity fragmentActivity, @
         if (fragment == null) {
             fragment = DefaultErrorDialogFragment.newInstance(errors);
         }
-        HyperwalletInsight.getInstance().trackError(fragmentActivity, pageName, pageGroup,
+
+        String errorMessage = errors.get(0).getMessage();
+        if (errorMessage == null || errorMessage.trim().length() == 0) {
+            errorMessage = ErrorUtils.getMessage(errors, fragmentActivity.getResources());
+        }
+        HyperwalletInsight.getInstance().trackError(fragmentActivity, pageName, pageGroup, null,
```
manikandan-altmtrk: Suggestion, I think it is not good practice to send value as **null**.

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/AddTransferMethodFragment.java
```diff
@@ -482,7 +482,7 @@ public void showInputErrors(@NonNull final List<HyperwalletError> errors) {
                             focusSet = true;
                         }
                         HyperwalletInsight.getInstance().trackError(context,
-                                TAG, PageGroups.TRANSFER_METHOD,
+                                TAG, PageGroups.TRANSFER_METHOD, null,
```
manikandan-altmtrk: Here also same. see above comments please if you have time.

commonui/src/main/java/com/hyperwallet/android/ui/common/insight/HyperwalletInsight.java
dnip-hw: I created `add()` and `addAll()` methods to ErrorParamsBuilder. So the user can optionally add the extra params. This will avoid the need to pass null.

PR #127, opened 2019-11-26
Task/hw 57735 input formatting

dnip-hw:
https://jira.site1.hyperwallet.local/browse/HW-57735


transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/PhoneWidget.java
```diff
@@ -83,8 +84,13 @@ public void beforeTextChanged(CharSequence s, int start, int count, int after) {
                 @Override
                 public void onTextChanged(CharSequence s, int start, int before, int count) {
                     if (before != count) {
-                        mValue = s.toString();
+                        mValue = formatToApi(s.toString());
                         mListener.saveTextChanged(getName(), getValue());
+                        if (mField.getMask() != null) {
```
fmattos-hw: I don't see this condition in the other widgets. Any specific reason?

transfermethodui/src/test/java/com/hyperwallet/android/ui/transfermethod/view/widget/AbstractMaskedInputWidgetTest.java
```diff
@@ -0,0 +1,95 @@
+package com.hyperwallet.android.ui.transfermethod.view.widget;
```
fmattos-hw: I'm not sure if this test should be here. I think it should be under the androidTest folder. Let's discuss this offline

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/AbstractMaskedInputWidget.java
```diff
@@ -0,0 +1,137 @@
...
+            }
+            if (textIndex >= apiValue.length() || patternIndex >= pattern.length()) {
+                break;
+            }
+        }
+        return formattedValue.toString();
+    }
+
+    @VisibleForTesting
+    String formatForTesting(@NonNull final String apiValue, @NonNull final String pattern) {
```
fmattos-hw: in this case I'd just make ethe format method package/default and remove this method

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/PhoneWidget.java
dnip-hw: oops, forgot to remove it. Realized the if-statement is not necessary since `formatToDisplay()` already does this null check

transfermethodui/src/test/java/com/hyperwallet/android/ui/transfermethod/view/widget/AbstractMaskedInputWidgetTest.java
dnip-hw: This is just a roboelectric test though? (actually, Roboelectric isn't even needed for this)

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/AbstractMaskedInputWidget.java
```diff
@@ -0,0 +1,137 @@
...
+    public AbstractMaskedInputWidget(@Nullable HyperwalletField field, @NonNull WidgetEventListener listener,
+            @Nullable String defaultValue, @NonNull View defaultFocusView) {
+        super(field, listener, defaultValue, defaultFocusView);
+    }
+
+    /**
+     * Removes the characters matching in the field's scrubRegex. This is to prepare the data for submission to the
+     * server.
+     *
+     * @param displayValue the value that would be coming from the component
```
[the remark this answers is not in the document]
dnip-hw: thanks, was having difficulty with some of the wordings

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/AbstractMaskedInputWidget.java
```diff
@@ -0,0 +1,137 @@
...
+        if (apiValue.length() == 0 || pattern.length() == 0) {
+            return "";
+        }
+        StringBuilder formattedValue = new StringBuilder();
+        String extraTokens = "";
+        int patternIndex = 0;
+        int textIndex = 0;
+
+        while (true) {
+            char token = pattern.charAt(patternIndex);
```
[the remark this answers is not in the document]
dnip-hw: not 100% sure what you mean, I'll just leave it for now to be safe.

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/AbstractMaskedInputWidget.java
```diff
@@ -0,0 +1,120 @@
...
+import com.hyperwallet.android.model.graphql.field.HyperwalletField;
+
+public abstract class AbstractMaskedInputWidget extends AbstractWidget {
+    private static final char NUMBER_TOKEN = '#';
+    private static final char TEXT_TOKEN = '@';
+    private static final char LETTER_OR_NUMBER_TOKEN = '*';
+
+    public AbstractMaskedInputWidget(@Nullable HyperwalletField field, @NonNull WidgetEventListener listener,
+            @Nullable String defaultValue, @NonNull View defaultFocusView) {
+        super(field, listener, defaultValue, defaultFocusView);
```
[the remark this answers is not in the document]
fmattos-hw: I don't think that's possible. The only reason I can think of to have it as nullable would be DefaultAccountWidget. But that class isn't being used and may never be

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/AbstractMaskedInputWidget.java
```diff
@@ -0,0 +1,143 @@
...
+        int textIndex = 0;
+
+        while (true) {
+            if (textIndex >= apiValue.length() || patternIndex >= pattern.length()) {
+                break;
+            }
+            char token = pattern.charAt(patternIndex);
+            char textChar = apiValue.charAt(textIndex);
+
+            if (token == '\\') {
```
fmattos-hw: why are we handling this use case? Is there a real case scenario where we'd have a backslash as part of the pattern? This seems pretty unrealistic to me. Let's discuss where this is coming from, cause I'm not convinced we should handle that scenario

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/PhoneWidget.java
```diff
@@ -68,23 +68,32 @@ public View getView(@NonNull final ViewGroup viewGroup) {
...
-                        mValue = ((EditText) v).getText().toString();
+                        String input = ((EditText) v).getText().toString();
+                        mValue = formatToApi(input);
                         mListener.valueChanged();
                     } else {
                         mListener.widgetFocused(PhoneWidget.this.getName());
                     }
                 }
             });
             editText.addTextChangedListener(new TextWatcher() {
```
fmattos-hw: Is this TextWatcher the same to all subclasses of AbstractMaskedInputWidget? If so, please move it there

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/NumberWidget.java
```diff
@@ -67,23 +67,32 @@ public View getView(@NonNull final ViewGroup viewGroup) {
...
 
                 @Override
                 public void onTextChanged(CharSequence s, int start, int before, int count) {
-                    if (before != count) {
-                        mValue = s.toString();
+                    if (!ignoreChange && before != count) {
+                        mValue = formatToApi(s.toString());
                         mListener.saveTextChanged(getName(), getValue());
+
+                        ignoreChange = true;
```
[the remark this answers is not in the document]
dnip-hw: We cannot set the text in afterTextChanged, it will result in an infinite loop of events.

I tracked the number of times the if-statement got triggered for pattern `1##A@@`, I entered 4 characters: 23bc (so that would result in 123Abc
- without ignoreChange boolean: 6 executions
- with ignoreChange boolean: 4 executions

From a functionality point of view, it seems exactly the same. It's sort like doing a preventDefault

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/PhoneWidget.java
dnip-hw: TextWidget, Numberwidget, PhoneWidget have their own private variables
```    private ViewGroup mContainer;
    private TextInputLayout mTextInputLayout;
    private String mValue;```

I'd have to move these to AbstractMaskedInputWidget as well. Do you want to do that?

File: transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/AbstractMaskedInputWidget.java (hunk 7 above)
Comment (fmattos-hw) [reply transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/AbstractMaskedInputWidget.java]: ok In that case, please add the "\\" as part of the constants

File: transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/PhoneWidget.java (hunk 8 above)
Comment (fmattos-hw) [reply transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/PhoneWidget.java]: Please do it. You can move them to AbstractMaskedInputWidget and expose setters as package default (no modifier)

File: transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/AbstractMaskedInputWidget.java [hunk 10]
```diff
@@ -0,0 +1,181 @@
...
+                        }
+                        patternIndex++;
+                }
+            }
+        }
+        return formattedValue.toString();
+    }
+
+    class InputMaskTextWatcher implements TextWatcher {
+        EditText mEditText;
```
fmattos-hw: this can be private and final

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/AbstractMaskedInputWidget.java
```diff
@@ -0,0 +1,181 @@
...
+
+        InputMaskTextWatcher(EditText editText) {
```
fmattos-hw: let's keep it consistent and add @NonNull and final

PR #132, opened 2019-12-04
HW-58501-unit test for field format excel sheet of pattern update

manikandan-altmtrk:
unit test case for 105 by using JUnitParamsRunner with the help of csv.


transfermethodui/src/test/java/com/hyperwallet/android/ui/transfermethod/view/widget/AbstractMaskedInputWidgetTest.java
```diff
@@ -22,8 +27,11 @@ public AbstractMaskedInputWidgetTest() {
     }
 
     @Test
-    public void testFormatToDisplay_usingExcelData() {
-        // TODO for Mani
+    @FileParameters("src/test/resources/formatter_mask_test_scenarios.csv")
+    public void testFormatToDisplay_usingExcelData(String description, String pattern, String scenario,
+            String inputValue, String formattedValue) {
+        assertThat("Description: " + description + "; Scenario: " + scenario,
```
fmattos-hw: I think it would be useful to add the pattern too.
Screen Shot 2019-12-04 at 4 10 46 PM
Screen Shot 2019-12-04 at 4 11 03 PM

dnip-hw: Yeah as I said in Jira, some scenarios are not correct in the spreadsheet. The failures tend to contradict the actual outputs.

After reviewing, I think it would be better to do a parameterized test.  See the samples I created below.
As well, reading from file seemed to alter the value of how the backslash character is interpreted when compared to the Android App consuming the graphql response.

Such as `##\\####`, when debugging the true value:
In file read, it becomes: ##\\####
But normal Java String it becomes: ##\####

manikandan-altmtrk: added @fmattos-hw Thanks.
Yes i got @dnip-hw. So i updated existing test case. Also added new test case based on your idea. So you can choose any one from those two test cases(**testFormatToDisplay_usingExcelData**, **testFormatToDisplay_fromExcelData**). Because i implemented to reduce the number of lines. 
FYI, I also got some failed test cases after code update. Most of the failure is ending with special character for input value shorter length than the pattern.
Screenshot 2019-12-05 at 8 26 06 PM
Screenshot 2019-12-05 at 8 26 15 PM
Screenshot 2019-12-05 at 8 26 26 PM
Screenshot 2019-12-05 at 8 26 34 PM
Screenshot 2019-12-05 at 8 26 42 PM
Screenshot 2019-12-05 at 8 26 50 PM
Screenshot 2019-12-05 at 8 26 59 PM
Screenshot 2019-12-05 at 8 27 09 PM
Screenshot 2019-12-05 at 8 27 17 PM

dnip-hw: I checked all 9 failing cases, I would argue that the expected output from Excel is wrong. I corrected them to the expected value. I list them below and put it on the ticket:
```
// fail 1, row 22, excel is wrong (11#), shouldn't add suffix
// fail 2, row 47, excel is wrong (aa@), shouldn't add suffix
// fail 3, row 67, excel is wrong (11), user entered matching dash, we should enter it too
// fail 4, row 73, excel is wrong (11*), shouldn't add suffix
// fail 5, row 76, excel is wrong (11*-N), * character only accepts 0-9AZaz
// fail 6, row 77, excel is wrong (aa*-a), * character only accepts 0-9AZaz
// fail 7, row 87, excel is wrong (1a-), * character only accepts 0-9AZaz
// fail 8, row 99, excel is wrong (@), shouldn't add prefix if no valid characers inputted
// fail 9, row 100, excel is wrong (@a1a#a1a*), shouldn't add suffix```

PR #134, opened 2019-12-11
HW-54725 Selection Widget, prevent popup when using search. Fix rando…

dnip-hw:
https://jira.site1.hyperwallet.local/browse/HW-54725

- fix input select not closing after one click when search bar was enabled
- fix focus going to random spot after using the search bar


transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/AddTransferMethodFragment.java
```diff
@@ -502,6 +501,11 @@ public void showInputErrors(@NonNull final List<HyperwalletError> errors) {
         mPresenter.handleUnmappedFieldError(mWidgetInputStateHashMap, errors);
     }
 
+    @Override
+    public boolean isWidgetSelectionFragmentDialogOpen() {
+        return getFragmentManager().findFragmentByTag(WidgetSelectionDialogFragment.TAG) != null;
```
fmattos-hw: could you try the following instead?
```suggestion
    @Override
    public boolean isWidgetSelectionFragmentDialogOpen() {
        return !isVisible();
    }
```
The suggestion should have the same effect with the advantage of not relying on the fragment manager. A use case to test though is when an error dialog is being displayed

dnip-hw: It differs a little
1. the flash is still there, but it is less prominent
2. if you click on the search icon and click the X without typing anything, you need to click the X twice to close the window.

fmattos-hw: ok then let's stick with your implementation. Thanks for trying 👍

PR #137, opened 2019-12-12
HW-54817 bottom border select transfer update

peter-joseph:


commonui/src/main/java/com/hyperwallet/android/ui/common/view/HorizontalDividerItemDecorator.java
```diff
@@ -31,11 +31,14 @@
 public class HorizontalDividerItemDecorator extends RecyclerView.ItemDecoration {
 
     private static final short DEFAULT_PADDING = 0;
+    private static final short TOP_DIVIDER_PADDING = 30;
```
fmattos-hw: I'm not sure of the implications of hardcoding this value. Maybe we could try getting the value from a resource, e.g Resources.getSystem().getDimension()

peter-joseph: yep make sense moved the value to resource instead

commonui/src/main/java/com/hyperwallet/android/ui/common/view/HorizontalDividerItemDecorator.java
```diff
@@ -33,9 +33,13 @@
     private static final short DEFAULT_PADDING = 0;
 
     protected final Drawable mHorizontalItemDivider;
+    private final boolean mHasTopDivider;
+    private final float topDividerPadding;
```
manikandan-altmtrk: Suggestion: i think naming may be for this variable **mTopDividerPadding**.

peter-joseph: 👍 thanks did not see that!

commonui/src/main/java/com/hyperwallet/android/ui/common/view/HorizontalDividerItemDecorator.java
```diff
@@ -96,4 +103,9 @@ public void onDrawOver(@NonNull Canvas c, @NonNull RecyclerView parent, @NonNull
             mHorizontalItemDivider.draw(c);
         }
     }
+
+    private float getTopDividerPadding(@NonNull final Context context) {
+        return context.getResources().getDimension(R.dimen.item_horizontal_padding)
+                * context.getResources().getDisplayMetrics().density;
```
dnip-hw: could assign to variable to avoid calling getResources() twice. Not sure if that improves performance or not...
Resources resources = context.getResources();
resources.getDimension()...

peter-joseph: The method is a supplementary routine on `HorizontalDividerItemDecorator` construction that means this method is invoked once only during creation of this class, as well as during creation of the `RecyclerView` that is using this instance.

The same concept as why are we not saving `Context` context change in the lifecycle of the mobile device, that remains true with `context.Resource` so caching this objects to save routine calls is not ideal.

PR #138, opened 2019-12-13
HW-58541 valueChanged on widget now only validates single item traver…

dnip-hw:
https://jira.site1.hyperwallet.local/browse/HW-58541
- changed interface of valueChanged to add AbstractWidget as argument, so that we can validate a single item

Tested
- check to see Insights called trackError on the last focused item only
- check submit form validates the entire form

https://jira.site1.hyperwallet.local/browse/HW-54774
 - scroll to first error

https://jira.site1.hyperwallet.local/browse/HW-58840
 - country pop up fix for API <= 27


transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/AddTransferMethodFragment.java
```diff
@@ -598,50 +598,68 @@ private void setVisibleAndDisableFields() {
...
+                    valid = isWidgetItemValid((AbstractWidget) v.getTag());
+                } else {
+                    isWidgetItemValid((AbstractWidget) v.getTag());
                 }
             }
         }
         return valid && hasWidget;
     }
 
+    private boolean isWidgetItemValid(AbstractWidget widget) {
```
fmattos-hw: To keep it consistent with other methods, please add @NonNull and final

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/AddTransferMethodFragment.java
```diff
@@ -598,50 +598,68 @@ private void setVisibleAndDisableFields() {
         mCreateTransferMethodButton.setBackgroundColor(getResources().getColor(R.color.colorSecondaryDark));
     }
 
-    private boolean performValidation(boolean bypassFocusCheck) {
+    /**
+     * Use this when we want to perform validation on the entire form, typically used when about to submit the form.
+     */
+    private boolean performValidation() {
```
fmattos-hw: This method contains a few issues:

- The variable context doesn't seem to be used

- Cast being done for each call of isWidgetItemValid when the cast has been previously done. You can invoke isWidgetItemValid(widget) instead of isWidgetItemValid((AbstractWidget) v.getTag())

- The implementation is a bit confusing where invoke the isWidgetItemValid in the if and else branches

- The return order is confusing. `return valid && hasWidget` I think it should be hasWidget && isValid 

I think the following implementation would be a little bit clearer. You can swap the usage of  containsInvalidWidget to true and change the logic accordingly, but it's the same idea
```suggestion
    private boolean performValidation() {
        boolean isWidgetValid;
        boolean containsInvalidWidget = false;
        // this is added since some phones triggers the create button but the widgets are not yet initialized
        boolean hasWidget = false;
        for (int i = 0; i < mDynamicContainer.getChildCount(); i++) {
            View v = mDynamicContainer.getChildAt(i);
            if (v.getTag() instanceof AbstractWidget) {
                hasWidget = true;

                AbstractWidget widget = (AbstractWidget) v.getTag();
                WidgetInputState widgetInputState = mWidgetInputStateHashMap.get(widget.getName());
                widgetInputState.setValue(widget.getValue());

                isWidgetValid = isWidgetItemValid(widget);
                if (!isWidgetValid) {
                    containsInvalidWidget = true;
                }
            }
        }
        return hasWidget && !containsInvalidWidget;
    }
```

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/DateWidget.java
```diff
@@ -116,7 +116,7 @@ public void onUpdate(@Nullable final String selectedDate) {
             try {
                 mEditText.setText(mDateUtils.convertDateFromServerToWidgetFormat(selectedDate));
                 mListener.saveTextChanged(getName(), getValue());
-                mListener.valueChanged();
+                mListener.valueChanged(this);
```
fmattos-hw: Keep it consistent with the other classes. Stick with either <Class>.this or just this to all calls to valueChanged

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/AddTransferMethodFragment.java
dnip-hw: used your code suggestion with a few changes

fmattos-hw: Cool, code looks good! Just some nitpicking... the variable isWidgetValid could be declared inside the loop with the method call isWidgetItemValid and inside the if, you could avoid invoking requireContext().getResources() twice by assigning it to a variable.

transfermethodui/src/main/res/layout/fragment_add_transfer_method.xml
```diff
@@ -5,6 +5,7 @@
               android:layout_width="match_parent"
               android:layout_height="match_parent"
               android:orientation="vertical"
+    android:focusableInTouchMode="true"
```
dnip-hw: I definitely used reformat code, etc. here, but it came out like this 🤷‍♂

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/AddTransferMethodFragment.java
```diff
@@ -598,48 +602,75 @@ private void setVisibleAndDisableFields() {
...
+     * @return true if the input is valid
+     */
+    private boolean isWidgetItemValid(@NonNull final AbstractWidget widget) {
+        boolean valid = true;
+        Context context = requireContext();
+
+        WidgetInputState widgetInputState = mWidgetInputStateHashMap.get(widget.getName());
+        widgetInputState.setValue(widget.getValue());
+        if (widget.isValid()) {
+            if (!widgetInputState.hasApiError()) {
```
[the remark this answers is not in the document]
peter-joseph: yeah retain it, I just remembered why we did that, we don't want to clear the api error after submit probably you can make some comment in the code

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/AddTransferMethodFragment.java
```diff
@@ -598,48 +619,74 @@ private void setVisibleAndDisableFields() {
...
+    }
+
+    /**
+     * Use this to perform validation on a single widget item, typically used while the user is inputting data.
+     *
+     * @param widget the widget to validate
+     * @return true if the input is valid
+     */
+    private boolean isWidgetItemValid(@NonNull final AbstractWidget widget) {
+        boolean valid = true;
```
manikandan-altmtrk: Suggestion: It might be good if the name **isWidgetValid** like. LGTM.

PR #144, opened 2019-12-21
bugfix/HW-59327 keyboard input text control

peter-joseph:


transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/TextWidget.java
```diff
@@ -71,7 +73,27 @@ public void onFocusChange(View v, boolean hasFocus) {
                     }
                 }
             });
-            editText.addTextChangedListener(new InputMaskTextWatcher(editText));
+            editText.addTextChangedListener(new TextWatcher() {
```
fmattos-hw: @peter-joseph if this approach solves our issue, please create a static inner class that inherits from InputMaskTextWatcher

peter-joseph: yep, that works great too!

PR #145, opened 2019-12-23
HW-59339 Transfer All Funds Toggle Flickers after Creating a Transfer - fixed

manikandan-altmtrk:
I believe, Those enable and disable actions are used to restrict UI interaction of component(Switch) when some(backend) process ongoing. We can also restrict UI interaction by using _setClickable(false)_ instead of using _setEnabled(false)_ as per android documentation without UI changes. So it can avoid the Toggle flicker. I checked manually. Let me know if you want any corrections or alternative methods.


fmattos-hw:
@manikandan-altmtrk  you're correct to assume the idea is to block user interaction by disabling the UI elements. I also think your suggestion would give the user a better experience. With that said, please apply the changes to the other UI elements as well

manikandan-altmtrk:
> @manikandan-altmtrk you're correct to assume the idea is to block user interaction by disabling the UI elements. I also think your suggestion would give the user a better experience. With that said, please apply the changes to the other UI elements as well
> Thanks

FYI: I added some comments on JIRA.

PR #151, opened 2020-01-16
task/HW-60059-TimeZone-Adjustment-for-ListPrepaidCardReceipts-Test

vinotb-altmtrk:
task/HW-60059-TimeZone-Adjustment-for-ListPrepaidCardReceipts-Test


receiptui/src/androidTest/java/com/hyperwallet/android/ui/receipt/ListPrepaidCardReceiptsTest.java
```diff
@@ -308,7 +338,10 @@ public void testListPrepaidCardReceipt_checkDateTextOnLocaleChange() {
         onView(withId(R.id.list_receipts)).check(
                 matches(atPosition(0, hasDescendant(withText("- 8,90")))));
         onView(withId(R.id.list_receipts)).check(
-                matches(atPosition(0, hasDescendant(withText("1. Juni 2019")))));
+                matches(atPosition(0,
+                        hasDescendant(withText(formatDateTime(mActivityTestRule.getActivity().getApplicationContext(),
+                                DateUtils.fromDateTimeString("2019-06-01T22:49:17").getTime(),
```
manikandan-altmtrk: I think we are providing localisation for Date & time only. Do we going to implement localisation for content also?
Because it may not be look good if we implement localisation for date & time only i think. @fmattos-hw @peter-joseph

receiptui/src/androidTest/java/com/hyperwallet/android/ui/receipt/ListPrepaidCardReceiptsTest.java
```diff
@@ -377,7 +413,10 @@ public void testListPrepaidCardReceipt_clickTransactionDisplaysDetails() {
         onView(withId(R.id.transaction_title)).check(matches(withText(R.string.prepaid_card_sale)));
         onView(withId(R.id.transaction_amount)).check(matches(withText("- 10.00")));
         onView(withId(R.id.transaction_currency)).check(matches(withText("USD")));
-        onView(withId(R.id.transaction_date)).check(matches(withText("June 6, 2019")));
+        onView(withId(R.id.transaction_date)).check(
+                matches(withText(formatDateTime(mActivityTestRule.getActivity().getApplicationContext(),
+                        DateUtils.fromDateTimeString("2019-06-06T22:49:17").getTime(),
```
manikandan-altmtrk: It will be look good if we place the expected value(like June 6, 2019) instead of date & time.  I think we can achieve by this date format operation. Could you please give some opinion  @skoong .

receiptui/src/test/resources/prepaid_card_receipt_list_response.json
```diff
@@ -46,7 +46,7 @@
     {
       "journalId": "FISA_5240223",
       "type": "ADJUSTMENT",
-      "createdOn": "2019-03-31T23:55:17",
+      "createdOn": "2019-03-30T23:55:17",
```
skoong: Why is this changed?

receiptui/src/androidTest/java/com/hyperwallet/android/ui/receipt/ListPrepaidCardReceiptsTest.java
skoong: Yes, I agree that this is hard to read.  However we need to convert the date and time into the correct expected value which is in readable format (June 6, 2019) so not sure if thats doable here.
@vinotb-altmtrk I think you need to consider the timezone as well.

receiptui/src/test/resources/prepaid_card_receipt_list_response.json
vinotb-altmtrk: @skoong due to 31st, in IST when I try to run, it moved to April 1st and I see two receipts splits to april and march, and it fails the UI. hence I changed it.

receiptui/src/androidTest/java/com/hyperwallet/android/ui/receipt/ListPrepaidCardReceiptsTest.java
vinotb-altmtrk: @skoong Ok shyang. I will check whether if we can make readable format is possible.

receiptui/src/androidTest/java/com/hyperwallet/android/ui/receipt/ListPrepaidCardReceiptsTest.java
```diff
@@ -377,7 +404,9 @@ public void testListPrepaidCardReceipt_clickTransactionDisplaysDetails() {
         onView(withId(R.id.transaction_title)).check(matches(withText(R.string.prepaid_card_sale)));
         onView(withId(R.id.transaction_amount)).check(matches(withText("- 10.00")));
         onView(withId(R.id.transaction_currency)).check(matches(withText("USD")));
-        onView(withId(R.id.transaction_date)).check(matches(withText("June 6, 2019")));
+        onView(withId(R.id.transaction_date)).check(
+                matches(withText(dateConversion("June 6, 2019", "22:49:41", "MMMM d, yyyy",
```
manikandan-altmtrk: FYI: I can see the difference between time value and the mock json response file. i think it is not a major difference. But it a may produce test case fail. 
Could you please update this(Screen shot attached) code if your code works well?
Screenshot 2020-01-20 at 10 35 23 PM

receiptui/src/androidTest/java/com/hyperwallet/android/ui/receipt/ListPrepaidCardReceiptsTest.java
```diff
@@ -481,4 +512,17 @@ private void setLocale(Locale locale) {
         }
         resources.updateConfiguration(configuration, resources.getDisplayMetrics());
     }
-}
+
+    private String dateConversion(String expectedDate, String timeAdjustment, String dateFormat,
+            int flag) {
+        SimpleDateFormat simpleDateFormat = new SimpleDateFormat(dateFormat + "HH:mm:ss");
+        simpleDateFormat.setTimeZone(TimeZone.getTimeZone("GMT"));
```
manikandan-altmtrk: Could you please confirm this hardcoded TIMEZONE with @skoong.

skoong: @vinotb-altmtrk This is incorrect, you need to convert the time which is in the JSON file (GMT) to the timezone which is in the device for display.  There is already something that does this in ListUserReceiptsTest.java:278.  You just need to generalize this function to make it work for the Month header as well.

skoong: You can leave the the input parameter as the date string, so it should look like this
dateConversion("2019-06-07T17:08:58", flags)

receiptui/src/test/resources/prepaid_card_receipt_list_response.json
skoong: Please change it back, once your timezone function should be able to handle this.

vinotb-altmtrk: Yes its done, and i changed back to 31st.

receiptui/src/androidTest/java/com/hyperwallet/android/ui/receipt/ListPrepaidCardReceiptsTest.java
vinotb-altmtrk: updated the date and time.

receiptui/src/androidTest/java/com/hyperwallet/android/ui/receipt/ListPrepaidCardReceiptsTest.java
vinotb-altmtrk: done.

testutils/src/main/java/com/hyperwallet/android/ui/testutils/util/DateConversion.java
```diff
@@ -0,0 +1,22 @@
...
+
+import static android.text.format.DateUtils.formatDateTime;
+
+import android.content.Context;
+
+import com.hyperwallet.android.ui.common.util.DateUtils;
+
+import java.util.Date;
+
+public class DateConversion {
```
skoong: This is in the wrong location, it should go into`hyperwallet-android-ui-sdk/testutils/src/main/java/com/hyperwallet/android/ui/testutils/util`.

skoong:
Please check why the following task is failing:
```
./gradlew clean lint testDebugUnitTest jacocoTestCoverageVerification
```

testutils/src/main/java/com/hyperwallet/android/ui/testutils/util/DateConversion.java
```diff
@@ -0,0 +1,22 @@
...
+
+    public static String convertDate(Context context, String dateString, int flag) {
```
skoong: We can move this file into `hyperwallet-android-ui-sdk/receiptui/src/androidTest/java/com/hyperwallet/android/ui/receipt`.

testutils/build.gradle
```diff
@@ -6,4 +6,5 @@ dependencies {
     implementation "androidx.test:rules:$testRulesVersion"
     implementation "androidx.test.espresso:espresso-core:$espressoVersion"
     implementation "com.squareup.okhttp3:mockwebserver:$mockServerVersion"
+    implementation project(':commonui')
```
skoong: I'd prefer not to add this dependency

testutils/src/main/java/com/hyperwallet/android/ui/testutils/util/DateConversion.java
```diff
@@ -0,0 +1,22 @@
...
+        Date date = DateUtils.fromDateTimeString(dateString);
+        return convertDate(context, date, flag);
```
skoong: See if ApplicationProvider.getApplicationContext() works instead of passing in context as a parameter.  I think we can also move the flags into this class and expose two methods. One for returning the date found in the receipt details and another for returning the date found in the receipt list.

fmattos-hw:
@skoong @vinotb-altmtrk I think a better approach would be setting a default locale in the tests. This can be achieved by either creating a rule or by setting it in the context. That would guarantee that we don't use the same method used to format the actual date, then to give us the expected value (android.text.format.DateUtils.formatDateTime)

testutils/src/main/java/com/hyperwallet/android/ui/testutils/util/DateConversion.java
vinotb-altmtrk: ok

testutils/build.gradle
vinotb-altmtrk: ok

receiptui/src/androidTest/java/com/hyperwallet/android/ui/receipt/ListPrepaidCardReceiptsTest.java
```diff
@@ -81,18 +82,22 @@ protected Intent getActivityIntent() {
...
+    private TimeZone mDefaultTimeZone;
 
     @Before
     public void setup() {
         mMockWebServer.mockResponse().withHttpResponseCode(HTTP_OK).withBody(sResourceManager
                 .getResourceContent("authentication_token_response.json")).mock();
 
+        mDefaultTimeZone = TimeZone.getDefault();
         setLocale(Locale.US);
+        TimeZone.setDefault(TimeZone.getTimeZone("PST"));
```
manikandan-altmtrk: I am okay with this kind of approach. But Here I think it may produce error for some time zone and value.
In this case activity will launch based on PST time zone. But still mock json response still GMT 0(there is no time zone defined. So default is GMT 0). I think We need to change mock response to PST time zone or here keep time zone as GMT 0

receiptui/src/androidTest/java/com/hyperwallet/android/ui/receipt/ListPrepaidCardReceiptsTest.java
```diff
@@ -81,18 +73,22 @@ protected Intent getActivityIntent() {
...
                 }
             };
+    private TimeZone mDefaultTimeZone;
 
     @Before
     public void setup() {
         mMockWebServer.mockResponse().withHttpResponseCode(HTTP_OK).withBody(sResourceManager
                 .getResourceContent("authentication_token_response.json")).mock();
 
+        mDefaultTimeZone = TimeZone.getDefault();
```
fmattos-hw: @skoong @vinotb-altmtrk if this starts to become more common in other tests, a test rule should be used. I'm ok with the current implementation as it's only in the receipt scope but if that becomes necessary in other places, use a rule

vinotb-altmtrk: @fmattos-hw Sure flavio. Noted.

receiptui/src/androidTest/java/com/hyperwallet/android/ui/receipt/ListPrepaidCardReceiptsTest.java
manikandan-altmtrk: @vinotb-altmtrk @skoong it only works when the time value below 8:30(24-hour format) from mock json. Because response is converting to USA/Pacific time from GMT. But the expected value still GMT. 
FYI: video attached in HW-60059.(I changed time for reproduce purpose)

skoong: We are hardcoding the expected result now, so in your video you should be checking that its June 5th as GMT is 8 hours ahead of PST.

PR #155, opened 2020-01-24
HW-60014 Schedule Transfer Activity result as Cancel(default) 

manikandan-altmtrk:
Removed  set result OK
Now result come as RESULT_CANCELED(0) when navigating back button clicked


shahardra:
How do I verify

PR #158, opened 2020-02-07
HW-59945 - progressbar issue fix. also progress bar color fix in list…

manikandan-altmtrk:
Progress bar issue fixed in create transfer, schedule transfer. Also fixed issue in receipt screen.
Details updated in Jira.


manikandan-altmtrk:
> Don't we need to set
> `android:layout_centerHorizontal="true" android:layout_centerVertical="true"` to align progress bar in center?
This one is needed to align the center when the immediate parent layout is Relative layout and we don't need full space of the parent layout.

PR #167, opened 2020-05-26
HW-63977- restrict copy and cut options ON Add transfer method screen

manikandan-altmtrk:
Restricted copy and cut options for sensitive data in addtransfermethod screen.

### DIFF
--- a/transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/ActionModeCallbackInterceptor.java
+++ b/transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/ActionModeCallbackInterceptor.java
@@ -8,12 +8,19 @@ public class ActionModeCallbackInterceptor implements ActionMode.Callback {
 
     @Override
     public boolean onCreateActionMode(ActionMode mode, Menu menu) {
-        return false;
+        return true;
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/widget/ActionModeCallbackInterceptor.java
```diff
@@ -0,0 +1,28 @@
...
+
+import android.view.ActionMode;
+import android.view.Menu;
+import android.view.MenuItem;
+
+public class ActionModeCallbackInterceptor implements ActionMode.Callback {
+
+    @Override
+    public boolean onCreateActionMode(ActionMode mode, Menu menu) {
+        return false;
```
jsaini-hw, line 11: We should still allow Paste. Only cut and copy is disabled.

manikandan-altmtrk: This code also allows paste when long press. However, I customized now with the paste options.

manikandan-altmtrk:
> What if you disable longClick? Will it still allow Paste and disable cut/copy/ select all?

Longclick will helpful to allow paste options when empty edit text field.

PR #168, opened 2020-06-02
HW-63528 Exported activity can potentially lead to Denial of Service

vwagh-hw:


transfermethodui/src/main/AndroidManifest.xml
```diff
@@ -15,7 +20,8 @@
         <activity
             android:name="com.hyperwallet.android.ui.transfermethod.view.SelectTransferMethodActivity"
             android:label="@string/title_activity_select_transfer_method"
-            android:theme="@style/AppTheme.NoActionBar">
+            android:theme="@style/AppTheme.NoActionBar"
+            android:permission="com.hyperwallet.android.ui.permission.SELECT_TRANSFER_METHOD">
```
jsaini-hw, line 24: Shouldn't ListTransferDestinationActivity also need this permission?

vwagh-hw: ListTransferDestinationActivity doesn't have intent-filter and so `android:exported` is false for it. Only Activity with `android:exported` as true can be launched by components of other app. And so it doesn't require permission protection.

https://developer.android.com/guide/topics/manifest/activity-element#exported

PR #174, opened 2020-07-09
feature/HW-64668 transfer method selection ux update

peter-joseph:


transfermethodui/src/main/res/values/strings.xml
```diff
@@ -28,6 +28,7 @@
     <string name="add_transfer_method_fee_label">Transaction Fee:</string>
     <string name="add_transfer_method_processing_time_label">Processing Time:</string>
 
+    <string name="feeAndProcessingTimeInformation">%s \u2022 %s</string>
```
jsaini-hw, line 31: there's fee label after fee value in the mockup `%s fee \u2022 %s`

peter-joseph: yep thanks 👍

PR #176, opened 2020-07-21
feature/HW-64925 add transfer ui update

peter-joseph:


transfermethodui/src/main/res/values/strings.xml
```diff
@@ -57,20 +55,19 @@
...
+    <string name="identification">IDENTIFICATION</string>
+    <string name="intermediary_account">INTERMEDIARY ACCOUNT</string>
+    <string name="mobileFeesAndProcessingTime">FEES AND PROCESSING TIME</string>
 
     <string name="search_hint">Search Hint</string>
     <string name="search_country">Search Country</string>
     <string name="search_currency">Search Currency</string>
     <string name="search_widget">Search</string>
 
-    <string name="account_information_section_header">Account Information - %s (%s)</string>
```
jsaini-hw, line 72: As per invision: Account Info still needs to have the format `ACCOUNT INFO: COUNTRY(CURRENCY)`

peter-joseph: hmm I thought we will follow content, will confirm in. channel

peter-joseph: updated based on decision in channel

PR #177, opened 2020-07-22
feature/HW-64673 manage transfer method ux update

peter-joseph:


transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/ListTransferMethodActivity.java
```diff
@@ -68,7 +70,7 @@ protected void onCreate(Bundle savedInstanceState) {
         setSupportActionBar(toolbar);
         getSupportActionBar().setDisplayHomeAsUpEnabled(true);
         getSupportActionBar().setDisplayShowHomeEnabled(true);
-        getSupportActionBar().setTitle(R.string.title_activity_list_transfer_method);
+        getSupportActionBar().setTitle(R.string.mobileTransferMethodsHeader);
```
jsaini-hw, line 73: I see 2 changes in this PR for header `title_activity_list_transfer_method` and `mobileTransferMethodsHeader`. Remove the one that's not used

commonui/src/main/res/values/strings.xml
```diff
@@ -31,4 +30,9 @@
     <string name="not_translated_in_braces">(not translated)</string>
 
     <string name="authentication_error_header">Authentication Error</string>
+
+    <string name="mobileTransferMethodsHeader">Transfer methods</string>
```
jsaini-hw, line 34: Ideally this should not be commonUI module but inside Transfer Method module

peter-joseph: yeah but it's used in transfers ui and transfer method account ui modules

transfermethodui/src/main/java/com/hyperwallet/android/ui/transfermethod/view/ListTransferMethodActivity.java
peter-joseph: `title_activity_list_transfer_method` this is just for manifest

peter-joseph: ok looks like duplicating the keys on every other module just works, removing the other keys and just use one single key `mobileTransferMethodsHeader`

PR #197, opened 2020-10-20
Task/dtserwfour 5 create transfer confirm details information ui tests

malaw-hw:


vwagh-hw (approved):
Approving, but since call to load destination were removed (seems like they were duplicate calls) please test manually to ensure functionality isn't broken

jsaini-hw (approved):

PR #274, opened 2021-07-07
Update SNAPSHOT

shanavas123:


build.gradle
```diff
@@ -24,7 +24,7 @@ allprojects {
 
     }
 
-    project.version = "1.0.0-beta11"
+    project.version = "1.0.0-beta12-SNAPSHOT"
```
vwagh-hw, line 27: don't we need to update core sdk SNAPSHOT here as well?

shanavas123: updated Core SDK version

aparthibanpaypal (approved):

